PR #3, opened 2020-03-05
Test out util change detection

jborean93:
##### SUMMARY
Testing change detection for C# module utils in a collection


jborean93:
Looks like this works, the utils are detected as changed and only it and it's users will run in the integration tests.

PR #40, opened 2020-03-21
win_service: feature bonanza

jborean93:
##### SUMMARY
Now that the module has been refactored with https://github.com/ansible-collections/ansible.windows/pull/34 we can start adding more features into it. This PR adds the following options to win_service

* error_control
* failure_actions
* failure_actions_on_non_crash_failure
* failure_command
* failure_reboot_msg
* failure_reset_period_sec
* load_order_group
* pre_shutdown_timeout_ms
* required_privileges
* service_type
* sid_info

With this change the only things missing from the win_service module are:

* Triggers
    * This is quite complex and potentially warrants being in its own module
* Security Descriptors
    * I'm hoping to create a generic framework for `win_<resource>_sd` modules that this will fall under
* Preferred nodes
    * Only valid for NUMA enabled hosts, cannot really test in CI
* Process protection level
    * Once you set this we cannot touch the service so it would be difficult to make idempotent as well as test in CI
* Mentioned by @briantist, ability to set env vars for a service process
    * I can't find any official docs from MS but it looks like this can be done in the registry https://stackoverflow.com/questions/305870/accessing-environment-variables-from-windows-services/2728574#2728574, not sure if we should expose this here or just recommend win_regedit  

Fixes: https://github.com/ansible-collections/ansible.windows/issues/11

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_service

### DIFF
--- a/plugins/modules/win_service.ps1
+++ b/plugins/modules/win_service.ps1
@@ -6,6 +6,8 @@
 #AnsibleRequires -CSharpUtil Ansible.Basic
 #AnsibleRequires -CSharpUtil ansible_collections.ansible.windows.plugins.module_utils.SCManager
 
+# Need to be able to specify values larger than [Int32]::MaxValue for certain time based options.
+$uint32Type = [Func[[Object], [UInt32]]]{ [UInt32]$args[0] }
 $spec = @{
     options = @{
         dependencies = @{ type = 'list'; elements = 'str' }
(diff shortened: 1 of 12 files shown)


briantist:
> * Mentioned by @briantist, ability to set env vars for a service process
>   
>   * I can't find any official docs from MS but it looks like this can be done in the registry https://stackoverflow.com/questions/305870/accessing-environment-variables-from-windows-services/2728574#2728574, not sure if we should expose this here or just recommend win_regedit

## Additional Background
The primary place this seems to keep popping lately is [around .NET Profilers](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/setting-up-a-profiling-environment). More and more APM and security products consist of or contain a profiler, and setting it requires environment variables. You can also only have 1 profiler in an application. 

This is where I found out about this, [troubleshooting NewRelic profiler conflicts with the vendor](https://docs.newrelic.com/docs/agents/net-agent/troubleshooting/profiler-conflicts).

[Looking at the Microsoft instructions for profiling a service](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/setting-up-a-profiling-environment#windows_service), you can see that the steps are very annoying: they want you to set a system environment variable, then restart the computer so that the SCM inherits the value, and then delete the system variable so that it doesn't effect _every other .net application on the system_.

So it seems in response most profiler developers are relying on setting the environment variables on individual services since it doesn't seem to require a reboot, just a restart of the individual service, and doesn't affect other services. 

Since a lot of these products are for web applications, this most often ends up being the Windows Process Activation Service (`WAS`) or sometimes World Wide Web Publishing Service (`W3SVC`).

Some known products setting env vars via registry:
- [NewRelic](https://docs.newrelic.com/docs/agents/net-agent/troubleshooting/profiler-conflicts)
- [DataDog](https://github.com/DataDog/dd-trace-dotnet/issues/222)
- [Stackify](https://docs.stackify.com/docs/prefix-system-configuration-changes)
- [Microsoft SCOM](https://ianblythmanagement.wordpress.com/2017/08/05/scom-2012-r2-agent-shows-apmapp-pool-crashes-in-2016-mg/)
- Azure Application Insights Status Monitor:
  - [One of their tests referencing the keys](https://github.com/microsoft/ApplicationInsights-dotnet/blob/35e4bb2624bcbe2af9fb5d1e724b28ffd49c460c/WEB/Test/E2ETests/TestApps/Net452/E2ETestApp/DockerFileStatusMonitor)
  - [A bug in the status monitor that left the key blank, maiming the services](http://apmtips.com/blog/2015/02/28/bug-with-statusmonitor-5-dot-0-uninstall/) -- we should keep this in mind if we implement to ensure we never leave the service with a blank key 😬 

The SCOM/App Insights ones are interesting in that while it is very difficult to find official documentation, at least 2 teams within Microsoft are shipping products that rely on the functionality.

## To Implement or Not
I think we should. 

* Using `win_regedit` is possible but awkward. What we really want to supply are environment variables, so key/value pairs. We can pass as a list with `win_regedit` if we pre-build the `{{ key }}={{ value}}` string for each one. But it would be much nicer to pass a `dict` imo.

* We would have to build the registry path by ourselves; it's not so bad, but ideally modules take care of that (this could be a possible enhancement to `win_service_info` either way; returning the registry key).

## Considerations
Applying the settings and checking for changes is very easy since it's just a single registry key.

Writing a CI test that just checks the registry value is easy.

I think it'd be beyond our scope but if we want to go as far checking the env vars of a running test service, then this is a little bit of a pain. [There's no API for getting the environment variables from another process](https://blog.gapotchenko.com/eazfuscator.net/reading-environment-variables). The author of that post has a published .NET component that can do it (I used it, it works). Or we could write a service that itself dumps its env vars.

----

This aside, looking forward to this PR and all the useful features it adds!

jborean93:
Thanks @briantist I'll merge this PR with the features as is but would be open to seeing someone implement the environment implementation if they wish to do so.

PR #44, opened 2020-04-02
Added quote filter and setup CI for unit tests

jborean93:
##### SUMMARY
Added the `ansible.windows.quote` filter to quote arguments when calling executables or in a Windows shell like cmd or powershell. The existing `quote` filter in ansible-base is not up to the task due to the difference escaping rules.

This PR also adds the unit tests matrixes to CI and cleans up some extra dirs that aren't needed.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
filters/quote


plugins/filter/quote.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def _quote_pwsh(s):
+    # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-5.1
+    if not s:
+        return "''"
+
+    # We should always quote values in PowerShell as it has conflicting rules where strings can and can't be quoted.
+    # This means we quote the entire arg with single quotes and escape the single quotes already in the arg.
+    return "'{0}'".format(s.replace("'", "''"))
```
briantist, line 63: @jborean93 I know this technique works most of the time and has been used elsewhere in ansible (I forget if it's in the powershell shell plugin or directly win winrm/psrp), but it is possible to "single" quote PowerShell strings with characters other than the standard apostrophe `'` .

In PowerShell it's easiest to handle this with `[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent()` but since that's  not available, we can probably just hardcode the chars.

The really interesting thing about it is that the character delimiting the start and end don't even have to be _the same character_ as long ad they are both valid "single quote" chars.

For example consider this:
```powershell
’abc'
```
That's a valid single quoted string. Even though:
```powershell
"'abc'"[0] -as [int]
"'abc'"[-1] -as [int]
```
The code gen function handles it correctly:
```powershell
[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent("'abc'")
```

The [source for the code gen function is here](https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/lang/codegen.cs#L21) but the really useful stuff for us is actually in the [CharExtensions class where they define the single quote characters](https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L18-L21) (these 4 are in addition to the standard apostrophe, [as seen in the `IsSingleQuote` function](https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L265-L272)).

The most likely place for this to bite someone is with non-English languages (this is actually how I found out about this, by mistake, forgetting to switch back to the English IME).

I think the key takeway from the code gen code is that those characters all need to be doubled against themselves, not changed to apostrophe. That prevents them from being used as string delimiters but ensures their content remains the same:
```powershell
[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent("won’t stop")
```

jborean93: Having a separate issue to track this would be great. I was unaware that chars other than `'` could be used as a quote and needed to doubled up in this scenario.

briantist: Yeah it really caught me by surprise too; I think if I wasn't working on a powershell project that dealt with text in multiple languages I never would have run into it.

Opened #45 to track.

Issue #45, opened 2020-04-05
PowerShell single quoted string can use a variety of characters


briantist:
##### SUMMARY

It is possible to "single" quote PowerShell strings with characters other than the standard apostrophe `'` . This comes up any place we want to generate some PowerShell and need to take unknown content and put it into a single quoted string (like the `quote` filter, or in connection/shell plugins).

Within PowerShell it's easiest to handle this type of thing with 
```powershell
[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent()
```
but since that's not available in Python, we can probably just hardcode the list of chars.

The really interesting thing about it is that the character delimiting the start and end don't even have to be _the same character_ as long as they are both valid "single quote" chars.

For example consider this:
```powershell
’abc'
```
That's a valid single quoted string. Even though:
```powershell
"’abc'"[0] -as [int]
"’abc'"[-1] -as [int]
```
Output:
```
8217
39
```

The code gen function handles it correctly:
```powershell
[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent("’abc'")
```
Output:
```
’’abc''
```

The [source for the code gen function is here](https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/lang/codegen.cs#L21) but the really useful stuff for us is actually in the [CharExtensions class where they define the single quote characters](https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L18-L21) (these 4 are in addition to the standard apostrophe, [as seen in the `IsSingleQuote` function](https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L265-L272)).

The most likely place for this to bite someone is with non-English languages (this is actually how I found out about this, by mistake, forgetting to switch back to the English IME).

I think the key takeway from the code gen code is that those characters all need to be doubled against themselves, not changed to apostrophe. That prevents them from being used as string delimiters but ensures their content remains the same:
```powershell
[System.Management.Automation.Language.CodeGeneration]::EscapeSingleQuotedStringContent("won’t stop")
```

---
The new quote filter sparked this originally:
[...]

PR #48, opened 2020-04-15
Fix up ps quoting escape rules

jborean93:
##### SUMMARY
When escaping quotes for a PowerShell argument we need to double up on more than just `'`. We also need to escape `'`, `’`, `‘`, `‚`, and `‛` by just repeating the char. Also fixes a bug when trying to escape `"` in a cmd argument where the escaped value need to be `\^"` and not just `^"`.

Fixes https://github.com/ansible-collections/ansible.windows/issues/45

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
quote

### DIFF
--- a/plugins/filter/quote.py
+++ b/plugins/filter/quote.py
@@ -8,15 +8,21 @@ import re
 
 from ansible.errors import AnsibleFilterError
 from ansible.module_utils.common.collections import is_sequence
+from ansible.module_utils._text import to_text
 
-_UNSAFE_C = re.compile(r'[\s\t"]')
-_UNSAFE_CMD = re.compile(r'[\s\(\)\%\!^\"\<\>\&\|]')
+_UNSAFE_C = re.compile(u'[\\s\t"]')
+_UNSAFE_CMD = re.compile(u'[\\s\\(\\)\\^\\|%!"<>&]')
+
+# PowerShell has 5 characters it uses as a single quote, we need to double up on all of them.
+# https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L265-L272
+# https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L18-L21
+_UNSAFE_PWSH = re.compile(u"(['\u2018\u2019\u201a\u201b])")
 
 
 def _quote_c(s):
     # https://docs.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way
     if not s:
-        return '""'
+        return u'""'
 
     if not _UNSAFE_C.search(s):
         return s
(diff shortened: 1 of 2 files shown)


jborean93:
Thanks @briantist for the pickup on this one. This should fix up the issue as well as fix a few encoding/cmd escaping edge cases I didn't test properly originally.

plugins/filter/quote.py
```diff
@@ -8,15 +8,21 @@
...
 
-_UNSAFE_C = re.compile(r'[\s\t"]')
-_UNSAFE_CMD = re.compile(r'[\s\(\)\%\!^\"\<\>\&\|]')
+_UNSAFE_C = re.compile(u'[\\s\t"]')
+_UNSAFE_CMD = re.compile(u'[\\s\\(\\)\\^\\|%!"<>&]')
+
+# PowerShell has 5 characters it uses as a single quote, we need to double up on all of them.
+# https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L265-L272
+# https://github.com/PowerShell/PowerShell/blob/b7cb335f03fe2992d0cbd61699de9d9aafa1d7c1/src/System.Management.Automation/engine/parser/CharTraits.cs#L18-L21
+_UNSAFE_PWSH = re.compile(u"(['\u2018\u2019\u201a\u201b])")
```
briantist, line 19: I might have missed it but doesn't this list also need to include the "standard" single quote/apostrophe `'` along with the extended 4?

plugins/filter/quote.py
```diff
@@ -46,21 +52,22 @@ def _quote_cmd(s):
...
     # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-5.1
     if not s:
-        return "''"
+        return u"''"
 
     # We should always quote values in PowerShell as it has conflicting rules where strings can and can't be quoted.
-    # This means we quote the entire arg with single quotes and escape the single quotes already in the arg.
-    return "'{0}'".format(s.replace("'", "''"))
+    # This means we quote the entire arg with single quotes and just double up on the single quote equivalent chars.
+    return u"'{0}'".format(_UNSAFE_PWSH.sub(u'\\1\\1', s))
```
briantist, line 70: Referring to my comment above, the character class would need single quote `'` so the sub catches it no?

plugins/filter/quote.py
jborean93: Yep, that's the first character in the regex group, after the `[`.

briantist: ha I swear I looked it over 15 times before posting and yeah my eyes just went right over it. I figured it must have been something like that since all the tests pass.

briantist:
#45 will be closed by this PR but lists at least 2 other instances like this in ansible/ansible. I can open sister issues there if you want.

jborean93:
Up to you, once this was in I was going to just open a PR there with these fixes

briantist:
> Up to you, once this was in I was going to just open a PR there with these fixes
works for me, no need to open a new one :)

jborean93:
ansible/ansible PR opened at https://github.com/ansible/ansible/pull/68968.

PR #53, opened 2020-05-15
Added Install From Media options

thlayli123:
Our enterprise uses the Install From Media option to promote our domain controllers. This saves time in an otherwise lengthy process. I am adding the information used by the IFM parameters to this and the powershell script.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

PR #54, opened 2020-05-15
Added Install From Media parameters

thlayli123:
Our enterprise uses the Install From Media option to promote our domain controllers. This saves time in an otherwise lengthy process. I am adding the information used by the IFM parameters to this and the powershell script.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


jborean93:
I would recommend changing the option name from `ifm_path` to something like `install_media_path` or `install_media_source`. That's closer to the underlying cmdlet parameters and a bit more descriptive than `ifm_path`. Whatever we decide on here should also be applied to https://github.com/ansible-collections/ansible.windows/pull/53.

thlayli123:
I have made the recommended changes. Changing to installation_media_path to make it easier to understand. I also made the change to 53

jborean93:
@thlayli123 `installation_media_path` is a bit of a handful, would you be ok with changing it to `install_media_path`? It also seems like this is only set on the module and not i the module docs or is that being done in #53 beacause it was an edit in GitHub?

Finally you will need to update https://github.com/ansible-collections/ansible.windows/pull/54/files#diff-30ecdffafdff5388eb23e9df6608199cR217 to reflect the new option.

thlayli123:
Thanks! I'll make the recommended change. I'm new with Git, so I appreciate your being patient with me :)

PR #61, opened 2020-05-30
win_domain and win_domain_controller necessary features

htaudah:
##### SUMMARY

A mismatch in the required features between win_domain and win_domain_controller will lead to unnecessary reboots/feature installations when both modules are used (which is the typical case). The comments show win_domain incorrectly assumed that AD-Domain-Services includes RSAT-AD-Tools and RSAT-AD-AdminCenter (not the case, at least for Win 2019). Now both will check for the same necessary features in a domain controller.

##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

win_domain and win_domain_controller

### DIFF
--- a/plugins/modules/win_domain.ps1
+++ b/plugins/modules/win_domain.ps1
@@ -9,13 +9,28 @@ $ErrorActionPreference = "Stop"
 
 # FUTURE: Consider action wrapper to manage reboots and credential changes
 
+$required_features = @("AD-Domain-Services","RSAT-ADDS", "RSAT-AD-AdminCenter")
+
+Function Get-MissingFeatures {
+    Write-DebugLog "Checking for missing Windows features..."
+
+    $features = @(Get-WindowsFeature $required_features)
+
+    If($features.Count -ne $required_features.Count) {
+        Throw "One or more Windows features required for a domain controller are unavailable"
+    }
+
+    $missing_features = @($features | Where-Object InstallState -ne Installed)
+
+    return ,$missing_features # no, the comma's not a typo- allows us to return an empty array
+}
+
 Function Install-Prereqs {
-    $gwf = Get-WindowsFeature AD-Domain-Services
-    if ($gwf.InstallState -ne "Installed") {
+    $missing_features = Get-MissingFeatures
+    if ($missing_features.Count -gt 0) {
         $result.changed = $true
 
-        # NOTE: AD-Domain-Services includes: RSAT-AD-AdminCenter, RSAT-AD-Powershell and RSAT-ADDS-Tools
-        $awf = Add-WindowsFeature AD-Domain-Services -WhatIf:$check_mode
+        $awf = Add-WindowsFeature $missing_features -WhatIf:$check_mode
         $result.reboot_required = $awf.RestartNeeded
         # FUTURE: Check if reboot necessary
 
--- a/plugins/modules/win_domain.ps1
+++ b/plugins/modules/win_domain.ps1
@@ -12,8 +12,6 @@ $ErrorActionPreference = "Stop"
 $required_features = @("AD-Domain-Services","RSAT-ADDS", "RSAT-AD-AdminCenter")
 
 Function Get-MissingFeatures {
-    Write-DebugLog "Checking for missing Windows features..."
-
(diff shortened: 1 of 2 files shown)


plugins/modules/win_domain.ps1
```diff
@@ -9,13 +9,26 @@ $ErrorActionPreference = "Stop"
 
 # FUTURE: Consider action wrapper to manage reboots and credential changes
 
+$required_features = @("AD-Domain-Services","RSAT-ADDS", "RSAT-AD-AdminCenter")
+
+Function Get-MissingFeatures {
+    $features = @(Get-WindowsFeature $required_features)
```
jborean93, line 15: Instead of referencing `$required_features` as a var outside of this function you should pass in a list of features where it is called.

plugins/modules/win_domain.ps1
```diff
@@ -9,13 +9,26 @@ $ErrorActionPreference = "Stop"
...
+
+    If($features.Count -ne $required_features.Count) {
+        Throw "One or more Windows features required for a domain controller are unavailable"
```
jborean93, line 18: Would be good to get the list of features that are missing to help with future debugging.

plugins/modules/win_domain.ps1
```diff
@@ -9,13 +9,26 @@ $ErrorActionPreference = "Stop"
...
+    $missing_features = @($features | Where-Object InstallState -ne Installed)
+
+    return ,$missing_features # no, the comma's not a typo- allows us to return an empty array
+}
+
 Function Install-Prereqs {
-    $gwf = Get-WindowsFeature AD-Domain-Services
-    if ($gwf.InstallState -ne "Installed") {
+    $missing_features = Get-MissingFeatures
+    if ($missing_features.Count -gt 0) {
```
jborean93, line 28: You can do this instead.

```suggestion
    if ($missing_features) {
```

By doing this you don't need the unary comma when returning/outputting the missing features in `Get-MissingFeatures` as the if statement will be true if it's a string or array. You could also just call `$missing_features = @(Get-MissingFeatures) to ensure the output is an array as well but that shouldn't be necessary.

plugins/modules/win_domain_controller.ps1
```diff
@@ -28,7 +28,7 @@ Function Write-DebugLog {
     }
 }
 
-$required_features = @("AD-Domain-Services","RSAT-ADDS")
+$required_features = @("AD-Domain-Services","RSAT-ADDS", "RSAT-AD-AdminCenter")
```
jborean93, line 31: What is the reasoning behind adding `RSAT-AD-AdminCenter`. We didn't do it before for `win_domain` and now it means we have an extra feature that will be added that may not be needed.

htaudah: I am fine with excluding it from both `win_domain` and `win_domain_controller` but a comment in `win_domain` (that I've removed) shows that the original author mistakenly thought it would be included with AD-Domain-Services:
https://github.com/ansible-collections/ansible.windows/blob/32bf7dd8744d4cdc55d1144983a4d35f364a74d8/plugins/modules/win_domain.ps1#L17-L18

plugins/modules/win_domain.ps1
htaudah: I find that things are more maintainable when object types don't change. Anyway, I've changed the condition as requested.

plugins/modules/win_domain_controller.ps1
htaudah: I've removed RSAT-AD-AdminCenter from both `win_domain` and `win_domain_controller`.

plugins/modules/win_domain.ps1
htaudah: Changed in both `win_domain` and `win_domain_controller`.

plugins/modules/win_domain.ps1
htaudah: Added the list of unavailable features to the thrown message.

jborean93:
Thanks for working on this @htaudah, just doing one last manual test to make sure things are still good as we don't run this in CI.

PR #63, opened 2020-06-03
win_acl - use path type for path option

briantist:
##### SUMMARY

it would be nice to be able to use environment variables directly in the path for `win_acl` rather than having to gather facts and use the env vars via templating.

The `win_acl_inheritance` module already does this.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_acl


briantist:
I see now this won't work with registry paths, so this would be a harder change. Closing this out for now.

PR #69, opened 2020-06-18
Fix more sanity issues and fix 2.11 sanity ignores

jborean93:
##### SUMMARY
Now that devel has been updated to 2.11 we need a ignore file for that version. Also take the chance to remove the rest of the sanity checks that we can.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sanity


plugins/modules/setup.ps1
```diff
@@ -58,6 +61,23 @@ Function Get-MachineSid {
     return $machine_sid
 }
 
+Function Get-WinRMCertificate {
+    <#
+    .SYNOPSIS
+    Gets all the certificates that are bound to a WinRM HTTPS listener.
+    #>
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSCustomUseLiteralPath", "",
+        Justification="Need to use widlcard in gci to get all the child listeners")]
```
mattclay, line 70:
```suggestion
        Justification="Need to use wildcard in gci to get all the child listeners")]
```

jborean93: Thanks

jborean93:
This fixes the ignored sanity tests that were failing before, there is a new changelog sanity test that will still need to be fixed in another PR.

Issue #31, opened 2020-03-11
setup: Windows gathering facts should honor gather_timeout


ronansalmon:
##### SUMMARY
Windows facts should honor gather_timeout

playbook:
```
- hosts: windows
  gather_facts: false
  roles:
    - gather_fact
```
role: gather_fact
```
- name: Getting hosts infos
  setup:
    gather_timeout: 3
```

gather_timeout is ignored. Setup module never gives up after 3 seconds on windows machines.

##### COMPONENT NAME
setup

mianos:
This change to winrm should prevent the infinite loop on locked files or corrupt registry.
https://github.com/diyan/pywinrm/pull/297

jborean93:
Unfortunately this isn't something that should be handled in pywinrm, we need something that is connection agnostic. It can easily be done in a module through PowerShell as we have something similar for async it just needs to be implemented.

mianos:
Do you have a suggestion as to how to handle the infinite loop waiting on the host, in this case, the ansible host? It is a pretty common occurrence when you have a non trivial number of windows hosts, daily for us.  It can't be easily handled in ansible as the pywinrm blocks indefinitely.

jborean93:
The infinite loop waiting is how WinRM works, it will continue to poll the host until the WSMan response indicates the process has finished. If something is still keeping it open on the remote side you need to determine what that is. This is somewhat documented in the [Receive processing rules for MS-WSMV](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wsmv/50cc2c38-e094-4d59-aec2-39e09f314862).

> A client SHOULD immediately issue a Receive message when a command is launched, whether or not it will be sending input using Send messages. To prevent deadlock, livelock, or time-out situations, the server can return Receive messages with empty string content, but typically it will delay responding until output is available, providing that wsman:OperationTimeout rules are not violated. If no output is available before the wsman:OperationTimeout expires, the server MUST return a WSManFault with the Code attribute equal to "2150858793". When the client receives this fault, it SHOULD issue another Receive request. The client SHOULD continue to issue Receive messages as soon as the previous ReceiveResponse has been received.

If you are experiencing this daily you will need to figure out

* Whether it blocks on a common scenario, i.e. running particular command
* Whether it always affects the same hosts
* If not, whether there is a common denominator between the hosts, i.e. OS version, PS version, etc
* Any event logs on the Windows host that indicate that something was killed or some other error
[...]

Issue #26, opened 2020-03-11
setup: Windows Setup wrong epoch timestamp (get-date prints epoch + local_tz)


jborean93:
_From @mkayontour on Sep 09, 2019 12:10_

##### SUMMARY
Running the module setup on Windows prints the wrong epoch timestamp. 
https://github.com/ansible/ansible/blob/a95f9c66401e99e72a12cb9d3a6d6c828bc1ecfe/lib/ansible/modules/windows/setup.ps1#L175 

```
PS C:\Users\ansible> get-date

Montag, 9. September 2019 14:06:33


PS C:\Users\ansible> get-date -UFormat %s
1568038002,68659
```

Windows prints the seconds since 1970 with the local timezone added. 

```
[vagrant@ansible ~]$ date
Mon Sep  9 12:07:14 UTC 2019
[vagrant@ansible ~]$ date +%s
1568030842
# epoch Timestamp from windows without nanoseconds
[vagrant@ansible ~]$ date -d @1568038002  
Mon Sep  9 14:06:42 UTC 2019
```

This issue is already known to powershell and it will be fixed in powershell version 6. 
See [Issue#2195](https://github.com/PowerShell/PowerShell/issues/2195)

In addition the nanoseconds aren't necessary to the epoch timestamp and so it shouldn't be used anymore. 

###### Possible Solution

To get the most accurate result as in the powershell bug described, get-date should be used like this. This is rounded and without nanoseconds 
`[int][double]::Parse((get-date ([DateTime]::UtcNow) -UFormat +%s))`

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
Module setup on windows 
##### ANSIBLE VERSION
Any

##### OS / ENVIRONMENT
Windows with Powershell version < 6.x 

_Copied from original issue: ansible/ansible#61990_

xenlo:
Hi @jborean93,

I give a try to your fix #73, but still get the same error: `msg: Invalid value for epoch value (1597253655.92744)` when I try to use it with `strftime`.

I tried by installing the 'ansible.windows' collection via `ansible-galaxy` but also by 'git cloning' the master branch into my virtualenv and get the same result.

Maybe it's due to some version stuff on the remote windows hosts?

jborean93:
So the format of the value has not changed with that PR but rather the timezone the value represents has. If you were having issues trying to filter it with `strftime` then you will continue to have issues after that change.

xenlo:
[...]

PR #78, opened 2020-06-27
Setup refactory

jborean93:
##### SUMMARY
This is a major refactor of the PowerShell `setup` module. On testing on a test VM, the time taken to run the full set of subsets has gone from 8.5 seconds down to 3.6 seconds.

It changes the following;

###### Breaking Changes

* Will now include the IPv6 scope on a link local address for `ansible_ip_addresses`
    * This was done so we are not omitting any information that might be useful
* The `ansible_processor` list will return the index before the name and manufacturer just like the POSIX facts
    * This was done so we align with the behaviour of the Python setup module
    * It is doubtful this value is used in any runtime logic
* The `ansible_date_time.epoch` will return the seconds since EPOCH in UTC like the POSIX module
    * This was done so we align with the behaviour of the Python setup module

###### Features

* Major refactor to speed up the process, limited the use of WMI where possible
* Implemented the `gather_timeout` option to restrict the time taken for an individual fact subset
* Added `ansible_architecture2` to be more like the POSIX fact

###### Bugfixes

* Remove the usage of WMI for a standard user to speed up the execution time and remove access denied errors.

Fixes https://github.com/ansible-collections/ansible.windows/issues/26
Fixes https://github.com/ansible-collections/ansible.windows/issues/29
Fixes https://github.com/ansible-collections/ansible.windows/issues/31
Fixes https://github.com/ansible-collections/ansible.windows/issues/42

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
setup


jborean93:
The sanity failures requires a fix in `ansible-test` due to the unique nature of the `setup.ps1` module https://github.com/ansible/ansible/pull/70376.

jborean93:
@briantist I am planning on merging this in soon before I create the first release of this collection. If you do fine any time it would be great if you could test out the changes here.

briantist:
> @briantist I am planning on merging this in soon before I create the first release of this collection. If you do fine any time it would be great if you could test out the changes here.

I blocked out some time time tomorrow afternoon (US Eastern) to check it out, looks really nice.

> The ansible_date_time.epoch will return the seconds since EPOCH in UTC like the POSIX module

This one I already know is going to hit us in a bunch of places; I have to look through them. Some are used as "unique" values so it probably won't matter, but others are being compared to local time output from other things.

In any case I support the breaking change; UTC makes more sense.

What do you think about adding a `local_epoch` as well? That would give a really straightforward replacement.

jborean93:
Having a local equivalent sounds fine to me. That was a hairy one where I wasn’t sure if we should really change it. I opted to do the breaking change because:

* the format of the value didn’t change, unlike the architecture one
* the source of the value didn’t change, unlike some of the other WMI changes I had to keep because I couldn’t find an alternative source

plugins/modules/setup.ps1
```diff
@@ -3,505 +3,1143 @@
...
+    @{
+        Subsets = 'date_time'
+        Code = {
+            $datetime = (Get-Date)
+            $datetimeUtc = $datetime.ToUniversalTime()
+
+            $ansibleFacts.ansible_date_time = @{
+                date = $datetime.ToString("yyyy-MM-dd")
+                day = $datetime.ToString("dd")
+                epoch = (Get-Date ($datetimeUtc) -UFormat '+%s')
```
briantist, line 594: Discussed in comments, but recommend `epoch_local` (or `local_epoch`, or whatever name) as a way to get at the old value easily.

jborean93:
Thanks for reviewing, weird that yours takes that long without a big reduction, I was finding it was a lot faster across the board on all my test hosts. I'm probably going to re-add the measuring times for each subsection so potentially we can find what the bottleneck is in your setup.

plugins/modules/setup.ps1
jborean93: Went with `epoch_local` in this case.

jborean93:
I've pushed a change that adds the `_measure_subset` option, are you able to run this one more time with that set to True? By doing that you get the `measure_info` facts back that tell us how long each subset took to run.

```json
"measure_info": {
    "all_ipv4_addresses, all_ipv6_addresses": 0.17186759999999998,
    "bios": 0.015571199999999999,
    "date_time": 0.015639,
    "distribution": 0.25006819999999996,
    "env": 0.01594,
    "facter": 0.09347119999999999,
    "interfaces": 0.17158779999999998,
    "local": 0,
    "memory": 0.20316209999999998,
    "platform": 0.4375382,
    "powershell_version": 0,
    "processor": 0.0781236,
    "uptime": 0.0156174,
    "user": 0.0156384,
    "virtual": 0.0156174,
    "windows_domain": 0.0156531,
    "winrm": 0.12498279999999999
},
```

Hopefully that can potentially tell us what may be running slower on your hosts.

briantist:
Sorry to add extra work for you! I can tell you that the time taken is probably mostly overhead; not ansible connections to Windows hosts take under ~8 seconds for any task. This is probably latency on my connection to the hosts, possibly made worse by using Kerberos as authentication method.

But that looks interesting going to try it out with `_measure_subset` now.

jborean93:
No worries, I'm actually in the middle of ripping out the runspace threading, it isn't adding any major benefit and probably adds more overhead that isn't needed.

briantist:
So here's the data (only using `psrp` this time):

* Single subset (`date_time` only) total task time 7.95s:
```json
"measure_info": {
    "date_time": 0.078124
}
```

* Full set 9.3s:
```json
"measure_info": {
    "all_ipv4_addresses, all_ipv6_addresses": 0.2795143,
    "bios": 0.0781275,
    "date_time": 0.1093798,
    "distribution": 0.6718657,
    "env": 0.1422718,
    "facter": 0.0781272,
    "interfaces": 0.2812267,
    "local": 0.0156266,
    "memory": 0.5312337,
    "platform": 1.062492,
    "powershell_version": 0,
    "processor": 0.1718491,
    "uptime": 0.0561348,
    "user": 0.0156268,
    "virtual": 0.015625399999999998,
    "windows_domain": 0.0159954,
    "winrm": 0.38933019999999996
}
```
Those add up to 3.9s, so quite respectable!

I guess it might be more surprising that the original one doesn't seem so slow in comparison, but we don't have the breakdown I guess.

For comparison, I also ran `win_ping` against the same host: 7.29s  (against 7.77s for single and 9.33s for full setup in this same run).

briantist:
> No worries, I'm actually in the middle of ripping out the runspace threading, it isn't adding any major benefit and probably adds more overhead that isn't needed.

That's a good point, the times are so small individually already. Is the data in `measure_info` including the runspace overhead or excluding it?

jborean93:
Excluding, it's basically the time it takes to invoke the fact subset scriptblock. There's a problem with platform taking more than 10 seconds when an AWS vm starts up so I'm trying to solve that. I have a feeling the machine SID stuff is the culprit or maybe the reboot required check but I'm trying to narrow it down a bit.

briantist:
Interesting, how long after startup? All the machines I'm testing against are in AWS too

jborean93:
I think it might be more initializing the host when it is first built. I'm only seeing the problem when using ansible-test in CI for 2019. If I was to reboot the host then things are fine, it's just the once which makes things difficult to debug.

jborean93:
After extensive testing against multiple hosts I think this is ready for prime time. Thanks @briantist for testing out the changes here.

mickaeltr:
Hello,
We are now seeing the following WARNING on some servers:
```
TASK [Gathering Facts] *********************************************************
[WARNING]: Error when collecting facter facts: Join-Path : Impossible de lier
l'argument au paramètre « Path », car il s'agit d'une chaîne vide.  Au
caractère Ligne:4 : 47  +                 $facterPath = Join-Path -Path $_
-ChildPath facter.ex ...  +                                               ~~
+ CategoryInfo          : InvalidData : (:) [Join-Path],
ParameterBindingValidationException      + FullyQualifiedErrorId :   ParameterA
rgumentValidationErrorEmptyStringNotAllowed,Microsoft.PowerShell.Commands.JoinP
athCommand      à <ScriptBlock>, <Aucun fichier> : ligne 4  à <ScriptBlock>,
<Aucun fichier> : ligne 3
```
Do you think that could be related to that refactor?
Thanks

jborean93:
@mickaeltr thanks for sharing the issue, I've opened https://github.com/ansible-collections/ansible.windows/pull/160 which fixes the problem there introduced by this change. In the future it would be best to just log a new issue and then link to the PR you believe causes the problem. We don't always look back at old PRs and closed issues so can miss these comments from time to time.

PR #94, opened 2020-08-14
Add support for setting ACL rules on windows certificate store keys

Hakon:
##### SUMMARY

Added support via the win_acl module which now supports the Cert:\<store location>\<storename>\<thumbprint> paths.
The code for this is only available through the C api for CAPI (crypto api) and CNG (Crypto API: Next Generation).
The interface to CAPI and CNG is implemented as a module util called CertACLHelper.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_acl

### DIFF
--- a/plugins/module_utils/CertACLHelper.cs
+++ b/plugins/module_utils/CertACLHelper.cs
@@ -13,7 +13,7 @@ using System.Security.Principal;
 //TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAccessRightsHelper -TypeName CertAccessRightsHelper
 //TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAclHelper -TypeName CertAclHelper
 
-namespace ansible_collections.ansible.windows.plugins.module_utils.CertAclHelper
+namespace ansible_collections.ansible.windows.plugins.module_utils.CertACLHelper
 {
 
     public class CryptHandle : SafeHandleZeroOrMinusOneIsInvalid
--- a/plugins/module_utils/CertACLHelper.cs
+++ b/plugins/module_utils/CertACLHelper.cs
@@ -6,6 +6,13 @@ using System.Security.Cryptography.X509Certificates;
 using Microsoft.Win32.SafeHandles;
 using System.Security.Principal;
 
+
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CryptHandle -TypeName CryptHandle
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.SafeSecurityDescriptorPtr -TypeName SafeSecurityDescriptorPtr
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAccessRights -TypeName CertAccessRights
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAccessRightsHelper -TypeName CertAccessRightsHelper
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAclHelper -TypeName CertAclHelper
+
 namespace ansible_collections.ansible.windows.plugins.module_utils.CertAclHelper
 {
 
--- a/plugins/module_utils/CertACLHelper.cs
+++ b/plugins/module_utils/CertACLHelper.cs
@@ -1,0 +1,275 @@
+using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using System.Security.AccessControl;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Win32.SafeHandles;
+using System.Security.Principal;
+
+namespace ansible_collections.ansible.windows.plugins.module_utils.CertAclHelper
+{
+
+    public class CryptHandle : SafeHandleZeroOrMinusOneIsInvalid
+    {
+
+        public CryptHandle()
+            : base(true)
+        {
+        }
+
+        public CryptHandle(IntPtr handle)
+            : base(true)
+        {
+            this.SetHandle(handle);
+        }
+
+        [DllImport("advapi32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        private static extern bool CryptReleaseContext(IntPtr safeProvHandle, uint dwFlags);
+
+        protected override bool ReleaseHandle()
+        {
+            return CryptReleaseContext(this.handle, 0);
+        }
+    }
+    public class SafeSecurityDescriptorPtr : SafeHandleZeroOrMinusOneIsInvalid
+    {
+        private int size = -1;
+
+        public SafeSecurityDescriptorPtr()
+            : base(true)
+        {
+        }
+
+        public SafeSecurityDescriptorPtr(uint size)
+            : base(true)
+        {
+            this.size = (int)size;
+            this.SetHandle(Marshal.AllocHGlobal(this.size));
+        }
+
+        public SafeSecurityDescriptorPtr(IntPtr handle)
+            : base(true)
+        {
+            this.SetHandle(handle);
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            try
+            {
+                Marshal.FreeHGlobal(this.handle);
+                return true;
+            }
+            catch
+            {
+                // semantics of this function are to never throw an exception so we must eat the underlying error and 
+                // return false. 
+                return false;
+            }
+        }
+    }
+    [Flags]
+    public enum CertAccessRights : uint
+    {
+        Read = 0x80000000,
+        FullControl = 0x10000000
+    }
+
+    public static class CertAccessRightsHelper
+    {
+        public static int ToAccessMask(this CertAccessRights accessRights)
+        {
+            return (int)accessRights;
+        }
+    }
+
+    public class CertAccessRule : AccessRule
+    {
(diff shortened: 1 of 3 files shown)


plugins/modules/win_acl.ps1
```diff
@@ -122,7 +127,7 @@ ElseIf ($null -eq $inherit) {
 
 # Bug in Set-Acl, Get-Acl where -LiteralPath only works for the Registry provider if the location is in that root
 # qualifier. We also don't have a qualifier for a network path so only change if not null
-if ($null -ne $path_qualifier) {
+if (($null -ne $path_qualifier) -and ($path_qualifier -ne "Cert:")) {
```
briantist, line 130: Is there a reason not to use the `Push-Location` semantics with the `Cert:` provider?
If so, I'd recommend explaining that in the comment above.

plugins/modules/win_acl.py
```diff
@@ -45,6 +45,7 @@
       FileSystemRights U(https://msdn.microsoft.com/en-us/library/system.security.accesscontrol.filesystemrights.aspx).
     - If C(path) is a registry key, rights can be any right under MSDN
       RegistryRights U(https://msdn.microsoft.com/en-us/library/system.security.accesscontrol.registryrights.aspx).
+    - If C(path) is a certificate key, rights can be Read and/or FullControl
```
briantist, line 48:
```suggestion
    - If I(path) is a certificate key, rights can be C(Read) and/or C(FullControl).
```

Hakon: Is the C -> I also wrong for line 46?

plugins/modules/win_acl.ps1
Hakon: The cert: provider does not use Set-Acl or Get-Acl, I will explain this in the comment.

Hakon:
Thanks for your review. I will write tests for the cert provider.

plugins/modules/win_acl.py
```diff
@@ -123,4 +124,12 @@
...
     type: deny
     state: present
+
+- name: Set certificate private key FullControl to Fed-Phil
+  ansible.windows.win_acl:
+    path: Cert:\LocalMachine\My\168ba8c488463f88c6648466a22484b6189e165f
+    user: Fed-Phil
+    type: allow
+    state: present
+    rights: FullControl
```
jborean93, line 134: Am I right in seeing that only `Read` and `FullControl` is implemented for certificate rights? They are the only ones implemented in the UI but it would be good for our docs to state that they are the only ones valid for certificate private keys.

plugins/modules/win_acl.py
```diff
@@ -45,6 +45,7 @@
       FileSystemRights U(https://msdn.microsoft.com/en-us/library/system.security.accesscontrol.filesystemrights.aspx).
     - If C(path) is a registry key, rights can be any right under MSDN
       RegistryRights U(https://msdn.microsoft.com/en-us/library/system.security.accesscontrol.registryrights.aspx).
+    - If I(path) is a certificate key, rights can be C(Read) and/or C(FullControl).
```
jborean93, line 48: Should state what version of the collection this was added in (1.1.0).

plugins/module_utils/CertACLHelper.cs
```diff
@@ -0,0 +1,282 @@
...
+                {
+                    byte[] sd = value.GetSecurityDescriptorBinaryForm();
+                    if (NCryptSetProperty(
+                        handle,
+                        "Security Descr",
+                        sd,
+                        (uint)sd.Length,
+                        SecurityInformationFlags.DACL_SECURITY_INFORMATION) != 0)
+                    {
+                        throw new Exception("Could not set security descriptor");
```
jborean93, line 268: We are throwing away the error code here, it seems like `NCryptSetProperty` returns the error code which we can pass into Win32Exception.

plugins/module_utils/CertACLHelper.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+    [Flags]
+    public enum CertAccessRights : uint
+    {
+        Read = 0x80000000,
+        FullControl = 0x10000000
+    }
+
+    public static class CertAccessRightsHelper
```
jborean93, line 86: It this necessary? Seems like we can just do `[int]$accessRights` in the PowerShell code and avoid having this altogether. Potentially you don't even need the cast in PowerShell as well but you would have to test that.

plugins/module_utils/CertACLHelper.cs
```diff
@@ -0,0 +1,282 @@
...
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CryptHandle -TypeName CryptHandle
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.SafeSecurityDescriptorPtr -TypeName SafeSecurityDescriptorPtr
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAccessRights -TypeName CertAccessRights
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAccessRightsHelper -TypeName CertAccessRightsHelper
+//TypeAccelerator -Name Ansible.Windows.CertAclHelper.CertAclHelper -TypeName CertAclHelper
+
+namespace ansible_collections.ansible.windows.plugins.module_utils.CertACLHelper
+{
+
+    public class CryptHandle : SafeHandleZeroOrMinusOneIsInvalid
```
jborean93, line 19: I feel like a few of these classes should be `private` to avoid having outside users actually depend on this code without creating an actual public API.

plugins/module_utils/CertACLHelper.cs
```diff
@@ -0,0 +1,282 @@
...
+
+                    if (NCryptGetProperty(
+                        handle,
+                        "Security Descr",
+                        securityDescriptorBuffer,
+                        securityDescriptorSize,
+                        ref securityDescriptorSize,
+                        SecurityInformationFlags.DACL_SECURITY_INFORMATION) != 0)
+                    {
+                        throw new Exception("Could not get security descriptor");
```
jborean93, line 219: Same here, I believe you can use the return value from `NCryptGetProperty` to derive the Win32 error message for the failure.

jborean93:
Sorry for the delay, the code seems pretty sound to me. We would definitely need to have some integration tests for this to make sure things don't break in the future. Feel free to should out if you need help with those.

The only other part I am slightly concerned about is how public the interfaces are in the C# code. It would be nice to start up by only making the classes that win_acl needs to use as `public` and the rest being `private`. This allows us to be more flexible in changing the code in the future if need be without worrying we've broken some 3rd party collection's usage of it.

opoplawski:
Can this reference certificates by name rather than thumbprint?

opoplawski:
Just commenting that we are using a locally patched version of this module to manage private key permissions and find it very helpful.  I hope that this can make it into the released version at some point.

Hakon:
I am sorry about not responding to this, I have had other things to worry about. I tried to write an integration test for it but I remember I ran into an issue with local test environment (I am developing this on a windows machine, might be the reason).
I am glad someone else found it useful even though I haven't had time to address these issues.

Is it just to write as part of the win_acl test playbook?

plugins/modules/win_acl.py
Hakon: So more than what is documented on line 48?

plugins/modules/win_acl.py
Hakon: Added comment about it being added in 1.8.0 (hopefully? 😄)

plugins/module_utils/CertACLHelper.cs
Hakon: I believe this was an issue on older powershell versions, but my memory may be wrong.

jborean93:
> I am sorry about not responding to this, I have had other things to worry about.

That's all good, nothing to apologise for, we all have stuff like this :)

> Is it just to write as part of the win_acl test playbook?

Pretty much, you will need code to generate the test pfx and import that in the store to apply the ACLs. Ideally you would have a step to verify the ACL that was set was done properly and handles things like check mode. The hardest part is probably creating the certs to test in integration target but that should be a simple `delegate_to: localhost` with openssl away. Feel free to shout out if you need anymore help with this.

specator:
PR was sadly closed.

@jborean93 
Please describe what kind of test you expect?

jborean93:
Tests to test out the very scenario this PR was adding to `win_acl`.

specator:
No, @jborean93, I'm just tying to understand what kind of tests you need.

E.g:
1. just simple play that will create cert, show current acls, apply acls, shows result. Can I use powershell (win_shell) for that tests
or I need to use smth else?
2. some kind of intergated tests that will run palys and check results automaticaly? Should I test that other parts work fine or there is other tests exist?

Where I can find guidelines for this =))

jborean93:
We have tests for this module in https://github.com/ansible-collections/ansible.windows/tree/main/tests/integration/targets/win_acl that is run during CI that will test out changes to this module. What needs to happen is more tests are added to https://github.com/ansible-collections/ansible.windows/blob/main/tests/integration/targets/win_acl/tasks/tests.yml that test out the functionality added. That is tests to set the ACL on a certificate store key and the assertions that go along with it. At a minimum we would need tests for:

* setting in check mode
* setting in normal mode
* setting idempotently

This is just an Ansible role so you would write it just like any tasks. You can use `win_command`/`win_shell`/`win_powershell` or whatever other module you need to do the setup as needed.

PR #113, opened 2020-10-05
win_environment - add a dictionary option to set multiple values at once

briantist:
##### SUMMARY

- Adds dict option `variables` so you can set multiple key/value pairs without looping over the module
- Previous `name` and `value` options still work
- Added tests for the new option
- Added a new return value for seeing the changes on all vars affected
- Existing return values still work correctly on single vars

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_environment

##### ADDITIONAL INFORMATION

Usage:
```yaml
- win_environment:
    state: present
    level: machine
    variables:
      VAR_ONE: value 1
      COOL_VAR_NAME: uncool value tho
      VAULT_ADDR: https://127.0.0.1
```

Internals:
- A function handles the setting and before/after/changed status of individual env vars
- Single name/value option set is implemented the same as passing in a single key/value dict to `variables` internally, which keeps testing relatively simple
- Fully backward compatible

### DIFF
--- a/plugins/modules/win_environment.ps1
+++ b/plugins/modules/win_environment.ps1
@@ -10,17 +10,14 @@ $spec = @{
     options = @{
         name = @{ type = "str" }
         level = @{ type = "str"; choices = "machine", "process", "user"; required = $true }
-        state = @{ type = "str"; choices = "absent", "present"; default = "present" }
+        state = @{ type = "str"; choices = "absent", "present" }
         value = @{ type = "str" }
         variables = @{ type = "dict" }
     }
-    required_if = @(
-        ,@("state", "present", @("value", "variables"), $true)
-        ,@("state", "absent", @("name"))
-    )
     mutually_exclusive = @(
         ,@("variables", "name")
         ,@("variables", "value")
+        ,@("variables", "state")
     )
     required_one_of = @(,@("name", "variables"))
     supports_check_mode = $true
@@ -46,12 +43,20 @@ function Set-EnvironmentVariableState {
         $Value,
 
         [Parameter()]
-        [ValidateSet('present', 'absent')]
         [String]
         $State
     )
 
     Process {
+        if (-not $State) {
+            $State = if (-not $Value) {
+                'absent'
+            }
+            else {
+                'present'
+            }
+        }
+
         $before_value = [Environment]::GetEnvironmentVariable($name, $Level)
 
         $ret = @{
--- a/plugins/modules/win_environment.ps1
+++ b/plugins/modules/win_environment.ps1
@@ -1,5 +1,6 @@
 #!powershell
 
+# Copyright: (c) 2020, Brian Scholer (@briantist)
 # Copyright: (c) 2015, Jon Hawkesworth (@jhawkesworth) <figs@unity.demon.co.uk>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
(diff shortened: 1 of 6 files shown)


plugins/modules/win_environment.ps1
```diff
@@ -1,61 +1,124 @@
...
+foreach ($kv in $envvars.GetEnumerator()) {
+    $name = $kv.Key
+    $value = $kv.Value
+
+    # When removing environment, set value to $null if set
+    if ($state -eq "absent" -and $value) {
+        $module.Warn("When removing environment variable '$name' it should not have a value '$value' set")
+        $value = $null
+    } elseif ($state -eq "present" -and (-not $value)) {
+        $module.FailJson("When state=present, value must be defined and not an empty string, if you wish to remove the envvar, set state=absent")
```
briantist, line 105: @jborean93 this is the part (from the original, just moved in this PR) that catches the case of `state=present` and `value=<null/empty>`

tests/integration/targets/win_environment/tasks/main.yml
```diff
@@ -1,21 +1,14 @@
...
 - name: fail to create environment value with null value
   win_environment:
     name: "{{test_environment_name}}"
     state: present
     level: machine
   register: create_fail_null
-  failed_when: 'create_fail_null.msg != "state is present but all of the following are missing: value"'
+  failed_when: 'create_fail_null.msg != "state is present but any of the following are missing: value, variables"'
 
 - name: fail to create environment value with empty value
```
briantist, line 13: @jborean93 this is the existing test case for the functionality preventing an accidental removal on `state=present` and `value=<null/empty>`

jborean93:
The changes seem good to me, the only thing that I think we should consider is making `variables` and `state` mutually exclusive. The end user can control the state by setting a value or leaving it null/empty string. That allows them to set and unset multiple variables in 1 go like

```
- win_environment:
    variables:
      key: value
      unset: ''
```

What are your thoughts on that?

nitzmahone:
+1 from me for @jborean93 's suggestion, since Windows doesn't "do" empty envvars, and I'd be surprised if they ever let it. If that happens, we could always revisit.

briantist:
> What are your thoughts on that?

It does a solve for multi-removals which can't be done as-is, so I like that it solves that imbalance.

My main concern about doing that is it brings back the possibility of accidentally unsetting things. I think Jon did a great job preventing that by forcing you to be explicit about when you want to remove.

For static values it's obvious when you're passing an empty string, I'm thinking more along the lines of accidental templating mistakes possibly erasing vars where currently that's not possible.

If you all think that's ok, I can go ahead and make that change. I think I'll have to revisit the implementation a little and add some new tests, but doesn't seem too bad (famous last words 😜)

jborean93:
I think it's ok, we can keep the warning when the `name/value/state` one is being used. I think if we want to enable this more efficient way we should support setting a value as null/"" to unset the var.

briantist:
@jborean93 @nitzmahone finally got around to making the requested changes, ready for review when you have a moment

jborean93 (approved):

jborean93:
Looks good, I just had to fix up the `version_added` field as the next version is 1.3.0 and one was in the wrong spot. We really should try and implement a sanity check for that at some point.

briantist:
> Looks good, I just had to fix up the `version_added` field as the next version is 1.3.0 and one was in the wrong spot. We really should try and implement a sanity check for that at some point.

@jborean93  Sanity check would be ideal, yeah. But it'd also be nice to point out in developer documentation, or the contributing section of the readme, how to find the next version of the collection. I never seem to get those right.

Issue #116, opened 2020-10-17
win_package: don't wait for uninstall to finish


inorangestylee:
##### SUMMARY
In my case im tried to install "FusionInventory agent" after removing of previously installed. For removing i use a win_package module with regedit provider and state absent but it's always returns me "changed: true" and don't uninstall software.

I'm tried to debug scripts (`win_package.ps1`) behavior and got a few guesses about a problem. I think that the root of the problem is in the Ansible.Process.cs. I watched the execution of the task in ansible and through the debug script launched in ISE. In ISE, the execution is successful and the software is removed (in step-by-step mode), but in ansible and ISE in realtime the execution is successful, and the software is not removed. This is due to the fact that the module erroneously returns the result until the moment the software is uninstalled, because uninstall.exe unpack itself in `C:\Users\username\AppData\Local\Temp\~nsuA.tmp` and starts a new process `Un_A.exe`, and `rc: 0` from the terminated uninstall.exe is returned to the task, after which the runspace is closed and the deletion does not continue. Verbose output look like where is 2 exit codes:
```
Stopped at: ConvertTo-Json -InputObject $args[0] -Depth 99 -Compress
[DBG]: PS C:\temp\ansible\debug>> 
{"changed":true,"invocation":{"module_args":{"force_basic_auth":false,"url_username":null,"client_cert_password":null,"creates_path":null,"password":null,"creates_version":null,"timeout":30,"use_default_credential":false,"proxy_password":null,"method":nu
ll,"log_path":null,"username":null,"headers":null,"use_proxy":true,"proxy_url":null,"provider":"registry","maximum_redirection":50,"path":null,"url_password":null,"proxy_username":null,"product_id":"FusionInventory-Agent","validate_certs":true,"proxy_use
_default_credential":false,"follow_redirects":"safe","expected_return_code":[0,3010],"chdir":null,"arguments":"/S","state":"absent","http_agent":"ansible-httpget","client_cert":null,"creates_service":null}},"reboot_required":false,"rc":0,"failed":false}

Stopped at: Set-Variable -Name LASTEXITCODE -Value $args[0] -Scope Global; exit $args[0]
[DBG]: PS C:\temp\ansible\debug>> 

Stopped at: Set-Variable -Name LASTEXITCODE -Value $args[0] -Scope Global; exit $args[0]
```
First from `Uninstall.exe` and second from target `Un_A.exe`

[...]

PR #117, opened 2020-10-22
Added Process util and add option to wait for child processes to be done for win_package

jborean93:
##### SUMMARY
Adds a slightly modified version of `Ansible.Process` and `Ansible.ModuleUtils.CommandUtil` to the collection. This util should be the one used for any new work going forward as it can include new features without waiting on ansible-base to be updated.

Adds the option `wait_for_children` to `win_package` that can be used to make sure any child processes spawned by the installer are finished before `win_package` returns. This is an opt-in behaviour as it could cause hangs in existing installers if they spawn any child processes that never exit.

Fixes https://github.com/ansible-collections/ansible.windows/issues/116

Need to add some tests for the new module_utils.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_package

### DIFF
--- a/plugins/module_utils/Process.cs
+++ b/plugins/module_utils/Process.cs
@@ -250,7 +250,7 @@ namespace ansible_collections.ansible.windows.plugins.module_utils.Process
         public Win32Exception(string message) : this(Marshal.GetLastWin32Error(), message) { }
         public Win32Exception(int errorCode, string message) : base(errorCode)
         {
-            _msg = String.Format("{0} ({1}, Win32ErrorCode {2})", message, base.Message, errorCode);
+            _msg = String.Format("{0} ({1}, Win32ErrorCode {2} - 0x{2:X8})", message, base.Message, errorCode);
         }
 
         public override string Message { get { return _msg; } }
@@ -271,7 +271,7 @@ namespace ansible_collections.ansible.windows.plugins.module_utils.Process
         /// </summary>
         /// <param name="lpCommandLine">The command line to parse</param>
         /// <returns>An array of arguments interpreted by Windows</returns>
-        public static string[] ParseCommandLine(string lpCommandLine)
+        public static string[] CommandLineToArgv(string lpCommandLine)
         {
             int numArgs;
             using (SafeMemoryBuffer buf = NativeMethods.CommandLineToArgvW(lpCommandLine, out numArgs))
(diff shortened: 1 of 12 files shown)


jborean93:
Changes are passing CI but I'm wondering whether I should strip out some of the logic in `Process.cs` and it really just a stub for calling the PInvoke APIs. The bulk of the logic for the process would go into `Process.psm1` and it would allow us to eventually make `Start-AnsibleWindowsProcess` more flexible and can mimic how `Start-Process` works. The down side is that any C# util that wants to use this code would have more work cut out to do so compared to what's available today.

briantist:
> Changes are passing CI but I'm wondering whether I should strip out some of the logic in `Process.cs` and it really just a stub for calling the PInvoke APIs. The bulk of the logic for the process would go into `Process.psm1` and it would allow us to eventually make `Start-AnsibleWindowsProcess` more flexible and can mimic how `Start-Process` works. The down side is that any C# util that wants to use this code would have more work cut out to do so compared to what's available today.

I think that direction ultimately is more useful, having more of the logic in the psm1.

> This util should be the one used for any new work going forward as it can include new features without waiting on ansible-base to be updated.

I like that this helps development move more quickly within the collections, seems a net positive.

Wondering how you see this affecting third party collections. Is there any concern that it'll be seen as a negative to take a dependency on a large-ish collection to use the module util, vs. using a util that's always in Ansible?

jborean93:
> I think that direction ultimately is more useful, having more of the logic in the psm1.

I am inclined to agree here but it's not anything we can't add at a future point in time. I could be here forever trying to replicate the cmdlets but for an MVP I think I'm happy with the current setup.

> Is there any concern that it'll be seen as a negative to take a dependency on a large-ish collection to use the module util, vs. using a util that's always in Ansible?

It is a slight concern but `ansible.windows` is seen as the base Windows collection. We could potentially look at moving things out into a "common/base" windows collection with just the utils and critical plugins but that can be done at anytime in the future with redirection. Until that is seen as a needed setup I'm definitely just wanting to keep things simple in the 2 Windows collections we have today.

As for the utils in `ansible/ansible`, they will probably be there for a while but once we've abstracted some of the execution logic into a subsystem like plugin interface the plan is to eventually rip that out as well and move it here, or into another collection. Considering they have been available in Ansible for a while now they are going to stay and hopefully unchanged. It's too inflexible to make a change there and then wait for a collection to increase the Ansible version but it's also not possible to rip it out of Ansible as they are key for things like become/async. So the end goal is to create an equivalent in a collection and hopefully encourage new development to target them going forward.

Issue #142, opened 2020-12-03
"unauthorized operation" error when installing package form CD-ROM (read only device)


ruzickap:
##### SUMMARY

I'm using the `win_package` to install the qemu-ga-x86_64.msi package the from CD-ROM (read-only) device:

```yaml
- name: Install Qemu Guest Agent (qemu-ga-x64.msi)
  win_package:
    path: "{{ virtio_win_iso_path }}\\guest-agent\\qemu-ga-x86_64.msi"
    creates_path: "{{ ansible_env['ProgramFiles'] }}\\qemu-ga"
```

Ansible is giving me this error:

```
    qemu: TASK [ansible-role-virtio-win : Install Qemu Guest Agent (qemu-ga-x64.msi)] ****
    qemu: task path: /Users/ruzickap/git/packer-templates/ansible/roles/ansible-role-virtio-win/tasks/virtio-win.yml:79
    qemu: Thursday 03 December 2020  06:59:47 +0100 (0:00:11.707)       0:03:07.773 *****
    qemu: redirecting (type: modules) ansible.builtin.win_package to ansible.windows.win_package
    qemu: Using module file /usr/local/Cellar/ansible/2.10.3_1/libexec/lib/python3.9/site-packages/ansible_collections/ansible/windows/plugins/modules/win_package.ps1
    qemu: Pipelining is enabled.
    qemu: <127.0.0.1> ESTABLISH WINRM CONNECTION FOR USER: vagrant on PORT 2507 TO 127.0.0.1
    qemu: EXEC (via pipeline wrapper)
    qemu: The full traceback is:
    qemu: Attempted to perform an unauthorized operation.
    qemu: At line:422 char:12
    qemu: +     $acl | Set-Acl -LiteralPath $path
    qemu: +            ~~~~~~~~~~~~~~~~~~~~~~~~~~
    qemu:     + CategoryInfo          : PermissionDenied: (E:\virtio-win\g...u-ga-x86_64.msi:String) [Set-Acl], UnauthorizedAccessException
    qemu:     + FullyQualifiedErrorId : System.UnauthorizedAccessException,Microsoft.PowerShell.Commands.SetAclCommand
    qemu:
    qemu: ScriptStackTrace:
    qemu: at Add-SystemReadAce, <No file>: line 422
    qemu: at <ScriptBlock>, <No file>: line 790
    qemu: at <ScriptBlock>, <No file>: line 1386
    qemu:
[...]

PR #147, opened 2020-12-07
win_package - fix installs from readonly location

jborean93:
##### SUMMARY
When installing from a readonly location the `win_package` module fails as it cannot add the ACE for the `SYSTEM` account to the path. This may not be necessary so we can just ignore a failure and wait until it either succeeds or fails later on.

Fixes https://github.com/ansible-collections/ansible.windows/issues/142

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_package


jborean93:
Confirmed to fix the original issue in https://github.com/ansible-collections/ansible.windows/issues/142#issuecomment-742287487.

Issue #118, opened 2020-10-28
win_service failing with permission issue in ansible 2.10.2 (works in 2.9.9)


Kollibri:
##### SUMMARY
I have a simple task that runs through a list of windows services and disables them.  It was working in ansible 2.9.9 without any issue, but in ansible 2.10.2 it is failing on two of the services with the following error:

Task:
```
- name: disable non-essential services
  win_service:
    name: "{{ item }}"
    state: stopped
    start_mode: disabled
  loop:
    - AxInstSV
    - AJRouter
    - ALG
    - bthserv
    - dmwappushservice
    - MapsBroker
    - lfsvc
    - SharedAccess
    - lltdsvc
    - wlidsvc
    - CDPSvc
    - NcbService
    - PhoneSvc
    - Spooler
    - PrintNotify
    - PcaSvc
    - QWAVE
    - RmSvc
    - SensorDataService
    - SensrSvc
    - SensorService
    - ShellHWDetection
    - ScDeviceEnum
    - SSDPSRV
    - WiaRpc
    - TabletInputService
    - upnphost
    - WalletService
    - Audiosrv
    - AudioEndpointBuilder
    - FrameServer
    - stisvc
    - wisvc
    - icssvc
    - WpnService
```

Error for the "RmSvc" service:
```
Unhandled exception while executing module: Cannot convert value "RmSvc" to type "ansible_collections.ansible.windows.plugins.module_utils.SCManager.Service". Error: "Failed to open service ''RmSvc'' (Access is denied, Win32ErrorCode 5 - 0x00000005)"
```

Error for the "ScDeviceEnum" service:
```
Unhandled exception while executing module: Cannot convert value "ScDeviceEnum" to type "ansible_collections.ansible.windows.plugins.module_utils.SCManager.Service". Error: "Failed to open service ''ScDeviceEnum'' (Access is denied, Win32ErrorCode 5 - 0x00000005)"
```

All the other services are disabled without issue.

If I roll back to ansible 2.9.9, it works as expected.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_service

##### ANSIBLE VERSION

```
ansible 2.10.2
  config file = /builder/.ansible.cfg
  configured module search path = ['/builder/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/site-packages/ansible
  executable location = /usr/local/bin/ansible
[...]

Issue #115, opened 2020-10-07
win_service fails with an existing windows service


rhjhunt:
##### SUMMARY

The module `win_service` fails to start a service when the service you are querying is a 'FILE_SYSTEM_DRIVER' and not a traditional service you would expect to find in the service manager.
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_service
##### ANSIBLE VERSION

```paste below
ansible 2.7.10
  config file = None
  configured module search path = ['/home/tagliabu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/tagliabu/ansible/ansible-env/lib/python3.6/site-packages/ansible
  executable location = /home/tagliabu/ansible/ansible-env/bin/ansible
  python version = 3.6.9 (default, Jul 17 2020, 12:50:27) [GCC 8.4.0]
```
**Note:** This was also verified with ansible 2.9, and ansible 2.10 using the `ansible.windows` collection.

##### OS / ENVIRONMENT

Windows 2016 virtual machine with vagrant

##### STEPS TO REPRODUCE

This is falling with Crowdstrke `csagent`, which is a 'FILE_SYSTEM_DRIVER' and not a traditional service. It seems like the `win_service` module is not able to find these types of service due to a quirk in how it finds the services installed.

```yaml
---
- name: Windows service testing
  hosts: all
  collections:
    - ansible.windows
  tasks:
    - name: Check Crowdstrike service
      ansible.windows.win_service_info:
        name: csagent

    - name: Start csagent
      ansible.windows.win_service:
        name: csagent
        state: started
```

##### EXPECTED RESULTS

To make sure that the Windows service is started.

##### ACTUAL RESULTS

The playbook fails with:

```
"Service 'csagent' is not installed, need to set 'path' to create a new service"
```

```paste below
TASK [crowdstrike-role : Check Crowdstrike service] ****************************
    task path: /mnt/c/Users/trabajo/ansible/roles/crowdstrike-role/tasks/install_windows.yml:49
    Using module file /home/tagliabu/ansible/ansible29-env/lib/python3.6/site-packages/ansible/modules/windows/win_command.ps1
    Pipelining is enabled.
    <127.0.0.1> ESTABLISH WINRM CONNECTION FOR USER: vagrant on PORT 55985 TO 127.0.0.1
    EXEC (via pipeline wrapper)
    changed: [windowsvm] => {
        "changed": true,
[...]

PR #149, opened 2020-12-07
win_service - Open drivers and relax required perms

jborean93:
##### SUMMARY
Specify explicit permissions when opening the service as we do not require all access. Also allow opening driver services which were being excluded by `Get-Service` due to how it was being called.

Fixes https://github.com/ansible-collections/ansible.windows/issues/115
Fixes https://github.com/ansible-collections/ansible.windows/issues/118

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_service
win_service_info


briantist (approved):
Looks good!

PR #160, opened 2020-12-15
setup - handle PATHs with blank entry

jborean93:
##### SUMMARY
When scaning the `PATH` env var we need to skip blank entries as it causes a binding issue when joining to `Join-Path -Path`.

Fixes https://github.com/ansible-collections/ansible.windows/pull/78#issuecomment-745229594

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
setup


mickaeltr:
Thanks @jborean93 for fixing it so fast. I was not sure if it was a bug in the code or on my server.
I eventually found out that it occurs when `PATH` ends with a `;`.

PR #189, opened 2021-03-01
Fix to support Windows Containers with win_package

tmeckel:
##### SUMMARY

This PR contains a small fix to `win_package.ps1` to successfully install packages (MSI) on Windows Containers.

Related: https://github.com/ansible-collections/community.docker/pull/96

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/plugins/modules/win_package.ps1
+++ b/plugins/modules/win_package.ps1
@@ -1401,7 +1401,7 @@ try {
             url { Get-UrlFile -Module $module -Url $path }
             unc { Copy-ItemWithCredential -Path $path -Destination $module.Tmpdir -Credential $credential }
         }
-        $path = $tempFile
+        $path = Resolve-Path -LiteralPath $tempFile
         $getParams.Path = $path
     } elseif ($path -and -not $pathType) {
         if (-not (Test-Path -LiteralPath $path)) {


briantist:
@tmeckel just curious, what's the error/problem when running in a container?
You'll need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) here too.

tmeckel:
@briantist that was a nasty one 😁  I had to create a Ansible role to deploy some stuff on Linux and Windows machines and obviously I used molecule to test my role. But after I got the Ansible Docker connection fixed to use Windows Containers the `win_package` module won't install anything and complained about that it could not find the file that I downloaded with `win_get_url`. Almost got frustrated until I discovered that this tiny change, that I contribute with this PR, fixed that issue. But please don't ask me why this only happened inside a Windows Container, because I simply don't know and to be honest I didn't have the time to analyze it.

tmeckel:
> 
> 
> @tmeckel just curious, what's the error/problem when running in a container?
> You'll need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) here too.

Regarding the changelog fragment: what "title" 😏 @felixfontein supplemented my changelog fragment with a change number and different title. So if you provide me this in beforehand I can provider the correct fragment upfront.

felixfontein:
@tmeckel the URL was the PR's URL, so something like `win_package - blabla (https://github.com/ansible-collections/ansible.windows/pull/189).`

jborean93:
I'm also curious as to what this is meant to fix as currently both functions should be returning the full path. Can you share your task that fails as well as the error message. It might indicate a problem with something else that should be fixed elsewhere instead of here.

Issue #168, opened 2021-01-05
win_user not idempotent when group name is capitalized differently


darthwalsh:
##### SUMMARY
Running `win_user` twice with `groups` with different name capitalization causes error.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_user

##### ANSIBLE VERSION
```paste below
ansible 2.10.3
  config file = /Users/walshca/adsk/ansible-automations/ansible.cfg
  configured module search path = ['/Users/walshca/adsk/ansible-automations/library']
  ansible python module location = /Users/walshca/adsk/ansible-automations/venv/lib/python3.9/site-packages/ansible
  executable location = /Users/walshca/adsk/ansible-automations/venv/bin/ansible
  python version = 3.9.0 (default, Oct 27 2020, 14:15:17) [Clang 12.0.0 (clang-1200.0.32.21)]
```

##### CONFIGURATION

```paste below
ANSIBLE_PIPELINING(/Users/walshca/adsk/ansible-automations/ansible.cfg) = True
ANSIBLE_SSH_ARGS(/Users/walshca/adsk/ansible-automations/ansible.cfg) = -o ForwardAgent=yes -o ControlMaster=auto -o ControlPersist=60s
DEFAULT_HOST_LIST(/Users/walshca/adsk/ansible-automations/ansible.cfg) = ['/Users/walshca/adsk/ansible-automations/inventory']
DEFAULT_MODULE_PATH(/Users/walshca/adsk/ansible-automations/ansible.cfg) = ['/Users/walshca/adsk/ansible-automations/library']
DEFAULT_ROLES_PATH(/Users/walshca/adsk/ansible-automations/ansible.cfg) = ['/Users/walshca/adsk/ansible-automations/roles']
DEFAULT_STDOUT_CALLBACK(/Users/walshca/adsk/ansible-automations/ansible.cfg) = debug
DEFAULT_VAULT_PASSWORD_FILE(/Users/walshca/adsk/ansible-automations/ansible.cfg) = /Users/walshca/adsk/ansible-automations/vault.password
HOST_KEY_CHECKING(/Users/walshca/adsk/ansible-automations/ansible.cfg) = False
INTERPRETER_PYTHON(/Users/walshca/adsk/ansible-automations/ansible.cfg) = auto_legacy_silent
```

##### OS / ENVIRONMENT
[...]

PR #191, opened 2021-03-08
win_user - use SIDs for group comparison

jborean93:
##### SUMMARY
Changes the internal group comparison logic to work with SIDs rather than the human name. Also exposes the ability to specify a group using it's SID which helps for localisation issues and the user wants to be more explicit.

Fixes https://github.com/ansible-collections/ansible.windows/issues/153
Fixes https://github.com/ansible-collections/ansible.windows/issues/168
Supersedes https://github.com/ansible-collections/ansible.windows/pull/154

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
win_user

### DIFF
--- a/changelogs/fragments/win_user-groups.yml
+++ b/changelogs/fragments/win_user-groups.yml
@@ -1,0 +1,5 @@
+minor_changes:
+- win_user - Support specifying groups using the SecurityIdentifier - https://github.com/ansible-collections/ansible.windows/issues/153
+
+bugfixes:
+- win_user - Compare existing vs desired groups in a case insenstive way - https://github.com/ansible-collections/ansible.windows/issues/168
--- a/plugins/modules/win_user.ps1
+++ b/plugins/modules/win_user.ps1
@@ -62,12 +62,26 @@ Function Get-AnsibleLocalGroup {
     [CmdletBinding()]
     param (
         [Parameter(Mandatory=$true)]
-        [String]
-        $Name
+        [string]
+        $Sid
     )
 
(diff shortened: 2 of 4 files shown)


jantari:
hey @jborean93 sorry for resurrecting, but I just recently realized this was addressed. Thank you very much.

Does this apply to `win_group_membership` as well? I can see the documentation for [`win_user`](https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_user_module.html#parameter-groups) was updated to explicitly mention the SID support, but it is not mentioned in the docs for [`win_group_membership`](https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_group_membership_module.html).

If SIDs are not supported there I can open a separate issue for that.

Thank you!

jborean93:
It compares the sids https://github.com/ansible-collections/ansible.windows/blob/main/plugins/modules/win_group_membership.ps1#L125-L153

jantari:
It looks like it *compares by SIDs* which addresses #168 but it looks up groups by name only:

https://github.com/ansible-collections/ansible.windows/blob/main/plugins/modules/win_group_membership.ps1#L110

so #153 is not possible with `win_group_membership`:

**Task**
```
- name: 'Grant RDP access for xyz'
  ansible.windows.win_group_membership:
    name: S-1-5-32-555
    state: present
    members:
      - 'domain\xyz-group'
```

**Results**
```
TASK [Grant RDP access for xyz] ********************************************************
fatal: [servername]: FAILED! => changed=false
  added: []
  msg: Could not find local group S-1-5-32-555
  name: S-1-5-32-555
```

PR #192, opened 2021-03-09
win_certificate_store - added ability to manage a service store

jborean93:
##### SUMMARY
Added the ability to manage the certificate store for a Windows service. Also made the `store_name` a free form field so it can manage 3rd party stores that are present.

I initially tried to allow creating and removing of custom stores in this module but the lack of documentation around moving a certificate store stopped me. I could remove the store once but it was never removed completely due to it being duplicated in various locations. This functionality will require more investigation before it can be implemented.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_certificate_store


jborean93:
I have tested the example with LDAPS to verify the key options and it works perfectly fine.

PR #197, opened 2021-03-19
Logiconly

placame:
##### SUMMARY
I've corrected the removing of rights, with post check for any remaining ACE and reporting if unsuccessful.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
win_acl

##### ADDITIONAL INFORMATION
When removing rights with state: absent, on several cases the result will be different than expected.
For example:
- Removing inherited rights: result show ok, not changed, while the inherited right is still there. Expected result failed, identity still has following rights: (list of rights with inheritance information)
- Removing part of explicit right:
  - Explanation: User: xy has **FullControl** rights, and removal of **Write** access is requested
  - Result: ok, not changed
  - Actual result in the system: the given identity still has the same explicit **FullControl** right
  - Expected result: user should not have **Write** access, but should have several other rights like **ReadAndExecute**, **Delete**, **ChangePermission** and so on...
- Removing a previously granted right:
  - Explanation: When an identity has **ReadAndExecute** right and an execution of this module grant successfully a **Write** right, then the following execution called to revoke the newly granted **Write** right.
  - Result: ok, not changed
  - Actual result in the system: the given identity still has **Write** right.
  - Expected result: revoking the previously granted **Write** right

For command output and further details see: https://github.com/ansible-collections/ansible.windows/pull/129


placame:
@jborean93 
I've opened the first pull request as requested regarding the logic correction.
Can you please help me regarding how to open the others?
I suppose if I open it to main, then it will create some conflicts as all want to modify the same files.

jborean93:
Lets just do them 1 at a time. You can open separate PRs if you wish to implement the stuff you want but you run the risk of having to deal with merge conflicts as we merge code in.

jborean93 (changes requested):
There are still a lot of changes here that are mostly stylistic changes or changes that aren't really needed which make this quite hard to review. Stuff like

* https://github.com/ansible-collections/ansible.windows/pull/197/files#diff-ebc9a4cdfaac4d872d5077dbe36791e64f096985dcd89b8a00f000e736a0778bR14-R17
* https://github.com/ansible-collections/ansible.windows/pull/197/files#diff-ebc9a4cdfaac4d872d5077dbe36791e64f096985dcd89b8a00f000e736a0778bR57-R59
* https://github.com/ansible-collections/ansible.windows/pull/197/files#diff-ebc9a4cdfaac4d872d5077dbe36791e64f096985dcd89b8a00f000e736a0778bR100

There's a whole bunch of stuff in the docs that seem to have changed like changing the `yes` -> `true` and removing some newlines in the examples. We want to keep all these things at a minimal to ensure we keep the git attribution and just make it easier to determine what was actually changed for this feature/bugfix vs other changes that aren't really important.

PR #200, opened 2021-03-29
win_template - Fix ansible internal call that changed slightly in devel

jborean93:
##### SUMMARY
The change in Ansible https://github.com/ansible/ansible/pull/73924 added a positional argument in between source and dest_path meaning the call in `win_template` was treating `dest` another another `source` option. Instead we use the kwarg to ensure any future additions won't affect the code here anymore. I checked both stable-2.10 and stable-2.9 to make sure `generate_ansible_template_vars` had the kwarg `dest_path` so this change will still work on those versions.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_template

### DIFF
--- a/changelogs/fragments/win_template-dest_path.yml
+++ b/changelogs/fragments/win_template-dest_path.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_template - Fix changed internal API that win_template uses to work with devel again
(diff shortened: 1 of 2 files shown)


dhabierre:
Hi.
This issue seems to be on the 2.10.16 too.
Can someone say if he can reproduce this behavior?
Thanks.

witokondoria:
It looks like ansible 2.10.16 is using ansible.windows 1.4.0 (which doesn't contain this fix), A manual bump of the collection was required on my setup (`ansible-galaxy collection install ansible.windows --upgrade`)

PR #210, opened 2021-04-14
win_powershell - more fixes

jborean93:
##### SUMMARY
Hopefully the last round of fixes for this module:

* In CI the tests are hanging at a random point with no indication of what has gone wrong. I'm hoping this fixes that but I need to run this multiple times to make sure
* Using `executable` left a whole bunch of temp files due to the use of `Add-Type` and some unexplained problem in the scope it was run
  * By using our `Add-CSharpType` we ensure that the proper temp dir is used which is explicitly cleaned up just in case
* When runnining something with `executable` we don't redirect the existing stdio handles so we don't have to reset them at the end
  * I doubt this was an issue but I'm trying anything here

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_powershell

### DIFF
--- a/plugins/modules/win_powershell.ps1
+++ b/plugins/modules/win_powershell.ps1
@@ -631,10 +631,21 @@ param (
 
     [Parameter(Mandatory=$true)]
     [String]
-    $SetScriptBlock
+    $SetScriptBlock,
+
+    [Parameter(Mandatory=$true)]
+    [String]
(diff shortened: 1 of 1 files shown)


jborean93:
I've rerun the CI tests about 5 times and cannot replicate the hangs from before so I'm hoping it is fixed.

PR #216, opened 2021-04-30
Updated certificate store name for intermediate CA

quatauta:
##### SUMMARY

Module `ansible.windows.win_certificate_store` documentation stated Windows certificate store name (`store_name: "CertificateAuthority"`) for intermediate certificate authorities.

Importing a certificate to store name `CertificateAuthority` raises error message `unable to find store ''CertificateAuthority'': (CertOpenStore failed (The system cannot find the file specified, Win32ErrorCode 2 - 0x00000002))`.

Using store name `CA` allowed me to import a certificate to certificate store for intermediate certificate authorities.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

Module `ansible.windows.win_certificate_store`

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/ansible.windows.win_certificate_store_module.rst
+++ b/docs/ansible.windows.win_certificate_store_module.rst
@@ -196,7 +196,7 @@ Parameters
                         <div>The store name to use when importing a certificate or searching for a certificate.</div>
                         <div><code>AddressBook</code>: The X.509 certificate store for other users</div>
                         <div><code>AuthRoot</code>: The X.509 certificate store for third-party certificate authorities (CAs)</div>
-                        <div><code>CertificateAuthority</code>: The X.509 certificate store for intermediate certificate authorities (CAs)</div>
+                        <div><code>CA</code>: The X.509 certificate store for intermediate certificate authorities (CAs)</div>
                         <div><code>Disallowed</code>: The X.509 certificate store for revoked certificates</div>
                         <div><code>My</code>: The X.509 certificate store for personal certificates</div>
                         <div><code>Root</code>: The X.509 certificate store for trusted root certificate authorities (CAs)</div>


jborean93:
I'll have to test this tomorrow as previously we just used the values from [StoreName](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.storename?view=net-5.0) to denote what store to open and that uses `CertificateAuthority`. Potentially the changes I've made in the refactor recently has broken this and we need to create a fix to allow both names for backwards compatibility.

Also just as an FYI documentation changes should be made int the relevant `<module>.py` file https://github.com/ansible-collections/ansible.windows/blob/main/plugins/modules/win_certificate_store.py instead of the rst docs. The rst files in `docs` are automatically generated from the module documentation around each release.

jborean93:
Yep that seems to be the case, https://github.com/ansible-collections/ansible.windows/pull/192 changed the way the module opened the certificate store away from one done by .NET to one where it calls the necessary Win32 APIs. We can see in the .NET reference code it manually maps the `StoreName` enum values to a string where `CertificateAuthority` becomes `CA` as you've noticed https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System/security/system/security/cryptography/x509/x509store.cs#L67-L91.

I'll update the module so that the old name will continue to work for backwards compatibility. Thanks for picking this up!

jborean93:
I've opened a PR to bring back the old behaviour https://github.com/ansible-collections/ansible.windows/pull/218, just to note `CA` will still continue to work going forward. The PR just makes sure that `store_name: CertificateAuthority` is translated to `CA` internally.

Issue #196, opened 2021-03-16
win_updates reboot flag functionality


idan-tankel:
##### SUMMARY
I faced with a strange issue when I was trying to install updates for Windows Servers
When a reboot required, the `reboot: yes` flag does not initiate a server restart

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

windows.win_updates
##### ANSIBLE VERSION

```paste below
ansible 2.9.14
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/home/polushin/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.5 (default, Jun 20 2019, 20:27:34) [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)
```

##### OS / ENVIRONMENT

Red Had Enterprise Linux 7.6, python 2.7

##### STEPS TO REPRODUCE

```yaml
 - name: Install updates
    win_updates:
      log_path: 'c:/ansible_wu.txt'
      reboot: yes
```

##### EXPECTED RESULTS

The module will initiate reboot as said in docs

##### ACTUAL RESULTS

The reboot is not initiated, and the playbook fails with a "reboot required to complete install update (Patch number)"
**Also, As I saw in the source code itself, the reboot ansible parameter is not referenced at all!**
It seems that the reboot "function" and  _reboot_  user parameter support was not created
Hope I'm not missing something :-)

Issue #193, opened 2021-03-09
Windows returning HRESULT: 0x8024043 resulting in flakiness when running the win_updates task


thehcma:
##### SUMMARY
The execution of the win_updates task (observed reliably when hitting a Windows 2012 R2 Standard - the same also occurs with Windows 2016 - version 1607) fails as follows:

```console
- msg: 'Failed to search for updates: Exception from HRESULT: 0x80240438'
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

The `win-updates`task

##### ANSIBLE VERSION

```conaole
  ansible 2.10.5
  config file = /ansible/1.0.0.197/ansible.cfg
  configured module search path = ['/users/ras_ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /ansible/1.0.0.197/venv/lib/python3.8/site-packages/ansible
  executable location = /ansible/1.0.0.197/venv/bin/ansible
  python version = 3.8.7 (default, Jan 11 2021, 19:39:01) [GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]
```

##### CONFIGURATION

```console
ACTION_WARNINGS(default) = True
AGNOSTIC_BECOME_PROMPT(default) = True
ALLOW_WORLD_READABLE_TMPFILES(default) = False
ANSIBLE_CONNECTION_PATH(default) = None
ANSIBLE_COW_PATH(default) = None
ANSIBLE_COW_SELECTION(default) = default
ANSIBLE_COW_WHITELIST(default) = ['bud-frogs', 'bunny', 'cheese', 'daemon', 'def
ANSIBLE_FORCE_COLOR(default) = False
ANSIBLE_NOCOLOR(default) = False
ANSIBLE_NOCOWS(default) = False
ANSIBLE_PIPELINING(default) = False
ANSIBLE_SSH_ARGS(default) = -C -o ControlMaster=auto -o ControlPersist=60s
ANSIBLE_SSH_CONTROL_PATH(default) = None
ANSIBLE_SSH_CONTROL_PATH_DIR(default) = ~/.ansible/cp
ANSIBLE_SSH_EXECUTABLE(default) = ssh
ANSIBLE_SSH_RETRIES(default) = 0
ANY_ERRORS_FATAL(default) = False
BECOME_ALLOW_SAME_USER(default) = False
[...]

Issue #143, opened 2020-12-03
win_update: WinRM failed to connect post reboot


Udayendu:
##### SUMMARY
**win_update** module is not responding properly to **reboot_timeout** and throwing the traceback even if the update & system reboot is successful.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
- win_update

##### ANSIBLE VERSION
```
pip3 show ansible
Name: ansible
Version: 2.10.1
Summary: Radically simple IT automation
Home-page: https://ansible.com/
Author: Ansible, Inc.
Author-email: info@ansible.com
License: GPLv3+
Location: /usr/local/lib/python3.8/dist-packages
Requires: ansible-base
Required-by:
```

##### OS / ENVIRONMENT
- Ubuntu 20.04 LTS [Ansible Provision Server]
- Windows 2016 Server [Guest OS]

##### STEPS TO REPRODUCE
```
- name: Install all security, critical, and rollup updates
  win_updates:
    category_names:
      - SecurityUpdates
      - CriticalUpdates
      - UpdateRollups
    reboot: yes
    state: installed
    reboot_timeout: 3600
```

##### EXPECTED RESULTS
Should complete the update successfully

##### ACTUAL RESULTS
Failed with the following traceback:

```
win_updates: attempting to get system boot time
<192.168.9.251> ESTABLISH WINRM CONNECTION FOR USER: redhatcloud\redhatadmin on PORT 5986 TO 192.168.9.251
EXEC (via pipeline wrapper)
win_updates: attempting to get system boot time
<192.168.9.251> ESTABLISH WINRM CONNECTION FOR USER: redhatcloud\redhatadmin on PORT 5986 TO 192.168.9.251
<192.168.9.251> ESTABLISH WINRM CONNECTION FOR USER: redhatcloud\redhatadmin on PORT 5986 TO 192.168.9.251
win_updates: attempting to get system boot time
<192.168.9.251> ESTABLISH WINRM CONNECTION FOR USER: redhatcloud\redhatadmin on PORT 5986 TO 192.168.9.251
EXEC (via pipeline wrapper)
<192.168.9.251> ESTABLISH WINRM CONNECTION FOR USER: redhatcloud\redhatadmin on PORT 5986 TO 192.168.9.251
win_updates: attempting post-reboot test command
EXEC (via pipeline wrapper)
win_updates: system successfully rebooted
win_updates: checking WUA is not busy with win_shell command
[...]

Issue #49, opened 2020-04-29
win_updates reboot:True - host is not rebooted, module returns error


DouglasHeriot:
##### SUMMARY

I've set `win_updates` `reboot: True` so that the hosts will be rebooted if necessary. However I've had one host fail with error message `A reboot is required before more updates can be installed`.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_updates

##### ANSIBLE VERSION

```paste below
ansible 2.9.7
  config file = /home/douglas/Code/infra-as-code/ansible.cfg
  configured module search path = ['/home/douglas/Code/infra-as-code/lib/ansible/modules']
  ansible python module location = /home/douglas/Code/infra-as-code/venv/lib/python3.6/site-packages/ansible
  executable location = /home/douglas/Code/infra-as-code/venv/bin/ansible
  python version = 3.6.9 (default, Apr 18 2020, 01:56:04) [GCC 8.4.0]
```

##### CONFIGURATION

```paste below
ANSIBLE_PIPELINING(/home/douglas/Code/infra-as-code/ansible.cfg) = True
COLLECTIONS_PATHS(/home/douglas/Code/infra-as-code/ansible.cfg) = ['/home/douglas/Code/infra-as-code/vendor']
DEFAULT_CLICONF_PLUGIN_PATH(/home/douglas/Code/infra-as-code/ansible.cfg) = ['/home/douglas/Code/infra-as-code/lib/ansible/plugins/cliconf
DEFAULT_FORKS(/home/douglas/Code/infra-as-code/ansible.cfg) = 30
DEFAULT_HOST_LIST(/home/douglas/Code/infra-as-code/ansible.cfg) = ['/home/douglas/Code/infra-as-code/inventory/staging.yml']
DEFAULT_INVENTORY_PLUGIN_PATH(/home/douglas/Code/infra-as-code/ansible.cfg) = ['/home/douglas/Code/infra-as-code/lib/ansible/plugins/inven
DEFAULT_LOOKUP_PLUGIN_PATH(/home/douglas/Code/infra-as-code/ansible.cfg) = ['/home/douglas/Code/infra-as-code/lib/ansible/plugins/lookup']
DEFAULT_MODULE_PATH(/home/douglas/Code/infra-as-code/ansible.cfg) = ['/home/douglas/Code/infra-as-code/lib/ansible/modules']
[...]

PR #225, opened 2021-05-21
Refactor win_updates

jborean93:
##### SUMMARY
This is a big refactor of `win_updates` to address some of the issues we have received over time. Some of the major changes being done here

* Removed the auto become work in the action plugin
  * The module now always runs the code as a scheduled task which is able to better survive WinRM problems
  * This means the `use_scheduled_task` option does nothing, the module is run as if it was set to `true`
* The module code kicks off the update to run in the background and returns immediate
  * The action plugin will repeatedly poll the update status in separated processes allowing it to better handle network disconnects
  * Async still runs as before but that's an edge case
* The download and install stage are run asynchronously and progress updates are returned back to the action plugin
  * Due to limitations when running with multiple hosts only the 25, 50, 75, 100% complete updates are displayed with `-vv`
  * The debug log will contain them all
* Improves the `log_path` entries to contain more information, including asynchronous progress update results
* Changes `whitelist`/`blacklist` to `accept_list` and `reject_list` to better match the new Ansible standard
  * The `filtered_reason` in the return result still states `whitelist`/`blacklist` due to backwards compatibility reasons
* Add a `hresult` to error message function in the action plugin to give users a better indication of why a failure occurred
  * These values and the message are taken from `wuerror.h` from the Windows SDK
  * Unfortunately there does not seem to be a way to do this programmatically in Windows so we are stuck with the manual mapping
* Probably so many other things I'm forgetting about

Fixes https://github.com/ansible-collections/ansible.windows/issues/13
Fixes https://github.com/ansible-collections/ansible.windows/issues/22
Fixes https://github.com/ansible-collections/ansible.windows/issues/25
Fixes https://github.com/ansible-collections/ansible.windows/issues/49
Fixes https://github.com/ansible-collections/ansible.windows/issues/143 - Or at least uses the new `win_reboot` code which better handles these situations
Fixes https://github.com/ansible-collections/ansible.windows/issues/193 - Or at least doesn't use become anymore which is the assumed problem
Fixes https://github.com/ansible-collections/ansible.windows/issues/196
Fixes https://github.com/ansible-collections/ansible.windows/issues/226

TODO:

- [x] Address `TODO` statements in the PR
- [x] Add unit tests with expected responses
- [x] Add changelog fragment
- [x] Test, test, and more tests

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
win_updates


jborean93:
I think the code and tests are mostly complete so the PR is ready for anyone wishing to try it out. I'm planning on doing some more indepth testing sometime next week to verify it has broken any scenarios I haven't encountered in testing.

PR #230, opened 2021-06-03
Enable unitcode testing in vscode

jborean93:
##### SUMMARY
Make it easy to run unit tests in vscode's IDE. The `.vscode/.env` and `.vscode/settings.json` are set to the default file that people can edit based on their environment setup without it being tracked for future revisions.

This also removes some extra code that was migrated originally and is either not used or no longer needed.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ansible-test


briantist:
thank you for posting this @jborean93 !

Issue #83, opened 2020-07-19
win_dsc: failed when used with psrp connection and Windows CJK locale


t-nakajimaa:
##### SUMMARY

win_dsc was failed when used with psrp connection and Windows  ja-JP locale.
I think it happens in CJK as well as  ja-JP.

I created the following dirty patch and applied it and it worked.

```bash
--- psrp.py     2020-07-06 07:36:27.355066100 +0000
+++ /usr/local/lib/python3.8/dist-packages/ansible/plugins/connection/psrp.py   2020-07-06 07:37:22.695066100 +0000
@@ -505,7 +505,7 @@ class Connection(ConnectionBase):
         if process.rc != 0:
             raise AnsibleError(to_native(process.stderr))

-        put_output = json.loads(process.stdout)
+        put_output = json.loads(process.stdout.decode('utf-8'))
         remote_sha1 = put_output.get("sha1")

         if not remote_sha1:
@@ -687,6 +687,7 @@ if ($bytes_read -gt 0) {
             negotiate_delegate=self._psrp_negotiate_delegate,
             negotiate_hostname_override=self._psrp_negotiate_hostname_override,
             negotiate_service=self._psrp_negotiate_service,
+            locale="ja-JP",
         )

         # Check if PSRP version supports newer read_timeout argument (needs pypsrp 0.3.0+)
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_dsc
psrp

##### ANSIBLE VERSION

```paste below
# ansible --version
ansible 2.9.10
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.8.2 (default, Apr 27 2020, 15:53:34) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below
empty
```

##### OS / ENVIRONMENT

target OS version: Windows Server 2019 version 1809 build 17763.379

```powershell
PS C:\Users\Administrator> Get-WinSystemLocale

LCID             Name             DisplayName
----             ----             -----------
1041             ja-JP            日本語 (日本)
```

##### STEPS TO REPRODUCE

```bash
# ansible-playbook -i hosts win_dsc.yml

PLAY [all] ********************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************
[...]

PR #244, opened 2021-07-09
win_dsc - fix locale problems

jborean93:
##### SUMMARY
The `PSDesiredStateConfiguration` fails to import when the UI culture doesn't match the installed UI culture. This works around this issue as best as possible by temporarily changing the current UI back and also not treat ever error as fatal.

Even if `$ErrorActionPreference = 'Continue'` masks a real fatal error (it shouldn't), the subsequent call to `Get-DscResource` will attempt to re-import it under the existing strict preference causing the same failure as today.

Fixes https://github.com/ansible-collections/ansible.windows/issues/83

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_dsc

### DIFF
--- a/changelogs/fragments/win_dsc-locale.yml
+++ b/changelogs/fragments/win_dsc-locale.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_dsc - Fix import errors when running against host that wasn't installed with the ``en-US`` locale - https://github.com/ansible-collections/ansible.windows/issues/83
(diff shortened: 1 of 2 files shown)


briantist (approved):
This is a great find, I get the feeling knowing this kind of thing about installed UI culture is going to come in handy in other places.

jborean93:
I feel like I've only just scratched the surface, was a fun one to track down.

PR #250, opened 2021-07-16
Fix typo

varnav:
`str` should be `string`, otherwise this error will occur:

```
value of type must be one of: none, binary, dword, expandstring, multistring, string, qword. Got no match for: str
```

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/ansible.windows.win_regedit_module.rst
+++ b/docs/ansible.windows.win_regedit_module.rst
@@ -289,7 +289,7 @@ Examples
         path: HKLM:\ANSIBLE\Control Panel\Mouse
         name: MouseTrails
         data: 10
-        type: str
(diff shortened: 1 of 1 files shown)


jborean93:
Thanks for the pickup, the module docs are generated from the Python documentation file so this change would have to be applied to https://github.com/ansible-collections/ansible.windows/blob/main/plugins/modules/win_regedit.py#L182 instead.

varnav:
Done: https://github.com/ansible-collections/ansible.windows/pull/251

PR #277, opened 2021-09-16
win_reboot - fix last_error raising

lydiym:
##### SUMMARY
Small fix for last_error raising in reboot util (python 3 compatibility)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
_reboot.py

##### ADDITIONAL INFORMATION
Original variable `e` is deleted outside the except block
https://stackoverflow.com/questions/24271752/except-clause-deletes-local-variable

Here's a simple example that reproduces the problem:
```python
def main():
    try:
        raise Exception("msg")
    except Exception as e:
        print(e)

    raise e


main()
```


jborean93:
Thanks for the pickup, I've pushed a commit that adds a changelog fragment but the rest looks great!

Issue #298, opened 2021-10-27
win_copy - recursive failure checking dest size with symlinks


jborean93:
##### SUMMARY
The `win_copy` module will continually try and drill into symlinks when checking the remote file size causing issues when there's a circular reference. The module should really ignore reparse points for the size check to simplify things as it is either going to point to a file/dir already being counted in the size check or somewhere outside dest which shouldn't be counted.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_copy

##### ANSIBLE VERSION
N/A

##### COLLECTION VERSION
N/A

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### STEPS TO REPRODUCE
This is most likely hard to reproduce as it requires symlinks in the `dest` folder to point to one of it's parent directories causing the recursion.

##### EXPECTED RESULTS
The module doesn't fail

##### ACTUAL RESULTS
The module eventually fails once it reaches a max recursion limit for symlinks as per https://github.com/ansible-collections/ansible.windows/pull/296.

 TASK [copy_task : Copy Tools Folder] ************************
 task path: /tmp/workspace/ansible/roles/tasks/some_tasks.yml:69
 Tuesday 26 October 2021  12:40:33 +0000 (0:01:32.132)       0:18:58.912 ******* 
 Tuesday 26 October 2021  12:40:33 +0000 (0:01:32.132)       0:18:58.911 ******* 
 redirecting (type: action) ansible.builtin.win_copy to ansible.windows.win_copy
 redirecting (type: action) ansible.builtin.win_copy to ansible.windows.win_copy
 Using module file /usr/local/lib/python3.8/dist-packages/ansible_collections/ansible/windows/plugins/modules/win_copy.ps1
 Pipelining is enabled.
 <target-server> ESTABLISH WINRM CONNECTION FOR USER: USER on PORT 5985 TO target-server
 EXEC (via pipeline wrapper)
 The full traceback is:
 Could not find a part of the path 'C:\Documents and Settings\.NET v2.0\AppData\Local\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data\Application Data'.
 At line:135 char:18
 + ...    $size = (Get-ChildItem -Literalpath $file.FullName -Recurse -Force ...
[...]

PR #304, opened 2021-11-09
win_copy - improve dest size check for weird edge cases

jborean93:
##### SUMMARY
Improve the error handling of the dest size check after copying a file remotely.

Fixes https://github.com/ansible-collections/ansible.windows/issues/298

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_copy

### DIFF
--- a/changelogs/fragments/win_copy-dest-calc-fix.yml
+++ b/changelogs/fragments/win_copy-dest-calc-fix.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_copy - improve dest folder size detection to handle broken and recursive symlinks as well as inaccesible folders - https://github.com/ansible-collections/ansible.windows/issues/298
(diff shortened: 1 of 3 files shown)


jborean93:
@michaelPf85 here is a further PR to improve the dest size handling to deal with recursive symlinks and inaccessible folders. While I was unable to reproduce your error I believe this should also solve that additional problem I was talking about in your PR.

michaelPf85:
That's great @jborean93 . I can confirm your code works when running in the context I was in the last time. I'm not sure how to really reproduce it because we were in a bug of win_copy. On top of that it was a pretty special case, because of "fake" user profiles created by IIS .NET App Pools.

jborean93:
No worries, thanks for testing anyway. Hopefully it should help in those rare edge cases.

PR #309, opened 2021-11-11
Simplify unit test runner

jborean93:
##### SUMMARY
Now that Ansible from a source checkout can be installed with `pip install -e .` we can simplify the logic in the collection loader plugin used in vscode.

Also adds PowerShell auto formatting to help make it easier to fix sanity problems.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ansible-test unit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -391,7 +391,3 @@ changelogs/.plugin-cache.yaml
 
 # ansible-test ignores
 tests/integration/inventory*
-
-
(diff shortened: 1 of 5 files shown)


jborean93:
Sanity failure is due to a recent change in devel and is fixed separately with https://github.com/ansible-collections/ansible.windows/pull/308.

Issue #310, opened 2021-11-13
Remote computer goes to sleep during win_updates being applied


agibson2:
##### SUMMARY
I am not sure if this is considered a bug or a feature request.  It appears that the win_updates process doesn't make sure the computer doesn't go to sleep during the updates.

This is with the latest version using pip and ansible-galaxy on CentOS 7 with RPM packages removed of ansible and using pip and ansible-galaxy to install ansible-core and ansible.windows. While doing updates, sometime in the middle or near the end, the computer went to sleep causing the updates to pause and giving an error for win_updates because the computer was asleep.

Error is below...

```TASK [Apply updates] ***************************************************************************************************
[WARNING]: Unknown failure when polling update result - attempting to cancel task: ('Connection aborted.', error(111,
'Connection refused'))
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ConnectionError: ('Connection aborted.', error(111, 'Connection refused'))
fatal: [test-pc2]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_updates

##### ANSIBLE VERSION

```
(pipansible) [root@Server-PC Ansible]# ansible --version
[DEPRECATION WARNING]: Ansible will require Python 3.8 or newer on the controller starting with Ansible 2.12. Current
version: 2.7.5 (default, Nov 16 2020, 22:23:17) [GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]. This feature will be removed
from ansible-core in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in
ansible.cfg.
/root/pipansible/lib/python2.7/site-packages/ansible/parsing/vault/__init__.py:44: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in the next release.
  from cryptography.exceptions import InvalidSignature
ansible [core 2.11.6]
  config file = /root/.ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /root/pipansible/lib/python2.7/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /root/pipansible/bin/ansible
[...]

PR #311, opened 2021-11-15
win_updates - prevent host from sleeping

jborean93:
##### SUMMARY
The module will not tell the host every minute not to go to sleep when running the long async tasks for Windows Updates.

Fixes https://github.com/ansible-collections/ansible.windows/issues/310

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_updates

### DIFF
--- a/changelogs/fragments/win_updates-sleep.yml
+++ b/changelogs/fragments/win_updates-sleep.yml
@@ -1,0 +1,4 @@
+bugfixes:
+- >-
+  win_updates - prevent the host from going to sleep if a low sleep timeout is set -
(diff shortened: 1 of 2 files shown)


agibson2:
I replaced the win_updates.ps1, uninstalled the last windows cumulative update for the win10 system, rebooted.  I then set the sleep time to 3 minutes.  Now the win10 system is in a constant reboot cycle trying to apply updates.  I logged in to make sure there was activity to make sure the reboot cycle wasn't related to a too quick sleep timeout but it still keeps getting rebooted.

This is the win_updates log from the win10 system....
https://pastebin.com/bGswAEx8

Issue #15, opened 2020-03-11
win_dsc fails to serialize properties into CimInstance


jborean93:
_From @Rukas on Aug 10, 2018 18:48_

##### SUMMARY

Using win_dsc results in the following error: `Failed to serialize properties into CimInstance.`

The same configuration in pure DSC works without issue:

```
        xIEEsc DisableIEEscAdmin {
            IsEnabled = $false
            UserRole  = "Administrators"   
        }
```

##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME

- win_dsc
##### ANSIBLE VERSION

```
2.6.2
```

##### CONFIGURATION

```
DEFAULT_FORKS(/opt/devops-ansible/playbooks/test/ansible.cfg) = 40
DEFAULT_HOST_LIST(/opt/devops-ansible/playbooks/test/ansible.cfg) = [u'/opt/devops-ansible/playbooks/test/hosts.yml']
DEFAULT_LOG_PATH(/opt/devops-ansible/playbooks/test/ansible.cfg) = /opt/devops-ansible/playbooks/test/playbook.log
PERSISTENT_CONNECT_TIMEOUT(/opt/devops-ansible/playbooks/test/ansible.cfg) = 30
```
##### OS / ENVIRONMENT

From Ubuntu 16.04.2
Managing Server 2016

##### STEPS TO REPRODUCE

hosts.yml
```yaml
sql:
  hosts:
    10.2.1.6:
      ansible_winrm_server_cert_validation: ignore
      ansible_connection: winrm
      ansible_user: "some user"
      ansible_password: "some vault password"
```
test_playbook.yml
```yaml
---
- hosts: sql
  gather_facts: true
  tasks:
    - name: Install PSDesiredStateConfiguration PowerShell Module
      win_psmodule:
        name: PSDesiredStateConfiguration
        state: present
    - name: Install xSystemSecurity PowerShell Module
      win_psmodule:
        name: xSystemSecurity
        state: present
    - name: Set IE security to disabled for Administrators
      win_dsc:
        resource_name: xIEEsc
        IsEnabled: false
        UserRole: 'Administrators'
        PsDscRunAsCredential_username: '{{ansible_user}}'
        PsDscRunAsCredential_password: '{{ansible_password}}'
```

##### EXPECTED RESULTS

The DSC configuration to be successful

##### ACTUAL RESULTS

```
ansible-playbook test_playbook.yml -i hosts.yml -vv --ask-vault-pass

[...]

PR #313, opened 2021-11-19
Check/Error for using a DSC composite resource

MKletz:
##### SUMMARY
Added a more descriptive error message for unsupported DSC resource types

relevant to #15 but does not fix

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_dsc

##### ADDITIONAL INFORMATION
per the discussion on #15 referencing these types of DSC resources causes the "Failed to serialize properties into CimInstance." which does not clearly explain the issue. I will be working on a PR to add support for these resources but this was a quick easy improvement to add in the mean time.


jborean93:
I'm not sure if we can easily solve the configuration issue with pslint sanity. We could technically install the OMI components that allow pwsh to parse the file into a MOF but that might be a bit too much bloat to add for a file used just for tests. My recommendation is to just skip this file explicitly in the sanity tests to enable CI to pass.

Basically you want to add the following to https://github.com/ansible-collections/ansible.windows/blob/main/tests/sanity/ignore-2.13.txt (we keep it in alphabetical order)

```
tests/integration/targets/win_dsc/files/xTestCompositeDsc/1.0.0/DSCResources/xTestComposite/xTestComposite.schema.psm1 pslint!skip # Pwsh cannot parse DSC to MOF on Linux
```

This will skip the PSScriptAnalyzer from running against this file entirely which means pwsh will no attempt to read it and convert it to a MOF file.

MKletz:
@jborean93 I believe it's having issues parsing the composite resource not a MOF. There is not a MOF being generated here to my knowledge.

jborean93:
> @jborean93 I believe it's having issues parsing the composite resource not a MOF

What I believe is happening is that when PowerShell parses the file mentioned it generates the MOF as part of that process. On Windows this is fine as the tools are builtin but on Linux/macOS it requires OMI to be present and we don't have that on our images. In the end skipping the `pslint` sanity test is fine for this file as it's just used for testing.

MKletz:
Gotcha, thanks for the help on the ignore. I agree it's not a huge deal.

Merge as you see fit, let me know if there's anything else that needs to be done.

jborean93 (commented):
Only thing missing is a changelog fragment. In this case I would say this is mostly a bugfix to produce a better error message so adding a file under https://github.com/ansible-collections/ansible.windows/tree/main/changelogs/fragments with something like should help

```yaml
bugfixes:
- win_dsc - Provide better error message when trying to invoke a composite DSC resource
```

Feel free to adjust the wording as you see fit.
```

jborean93 (approved):

MKletz:
done, will make sure the changelog fragment is done for future PRs.

PR #315, opened 2021-11-22
Error standardize

MKletz:
##### SUMMARY
Centralized all the error writing logic into the new function Write-AnsibleError

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_dsc

##### ADDITIONAL INFORMATION
The logic for raising an error back to ansible has been deduplicated with a new function

### DIFF
--- a/changelogs/fragments/win_dsc-write-error-standardize.yml
+++ b/changelogs/fragments/win_dsc-write-error-standardize.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- win_dsc - deduplicated error writing code with a new function. No error text was changed.
+- win_dsc - deduplicated error writing code with a new function. No actual error text was changed.
--- a/changelogs/fragments/win_dsc-write-error-standardize.yml
+++ b/changelogs/fragments/win_dsc-write-error-standardize.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- win_dsc - deduplicated error writing code with a new function. # no error text was changed just exit logic de-duplicated
+- win_dsc - deduplicated error writing code with a new function. No error text was changed.
--- a/changelogs/fragments/win_dsc-write-error-standardize.yml
+++ b/changelogs/fragments/win_dsc-write-error-standardize.yml
@@ -1,0 +1,2 @@
+minor changes:
+- win_dsc - deduplicated error writing code with a new function.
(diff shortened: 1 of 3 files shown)


plugins/modules/win_dsc.ps1
```diff
@@ -342,6 +342,32 @@ Function Invoke-SafeDscResource {
...
+    Error message text
+
+    .PARAMETER Failed
+    Whether or not the error triggers a failure. Defaults to '$true'.
+    #>
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory = $true)]
+        [String]$Msg,
+        [Boolean]$Failed = $true
```
jborean93, line 360: Being pedantic here this should really a switch, bool is useful on parameters if you have 3 states; true, false, and nothing whereas a switch is just on and off. But considering this is `Write-AnsibleError` then I don't see a need for having failed be optional at all here.

changelogs/fragments/win_dsc-write-error-standardize.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- win_dsc - deduplicated error writing code with a new function. # no error text was changed just exit logic de-duplicated
```
jborean93, line 2: Just as an FYI the comment here will no appear in the changelog, it will be lost when the changelog itself is generated on the next release.

plugins/modules/win_dsc.ps1
jborean93: Yea in this case there really isn't any reason, it can be added in the future if need be but these don't use `$module.FailJson/ExitJson` because the module object hasn't been initialised yet.

plugins/modules/win_dsc.ps1
```diff
@@ -342,6 +342,31 @@ Function Invoke-SafeDscResource {
...
 
+Function Write-AnsibleError {
+    <#
+    .SYNOPSIS
+    Raises a standard terminating error.
+
+    .PARAMETER Msg
+    Error message text
+
+    .PARAMETER Failed
```
jborean93, line 353: Failed is no longer a parameter now :)

MKletz: beat me to it lol

jborean93 (approved):

MKletz:
@jborean93 looks like a CI step is timing out

jborean93:
Yea Azure Pipelines is having troubles, it passed the most important steps though.

PR #320, opened 2021-12-13
win_updates - fix execution policy restrictions

jborean93:
##### SUMMARY
If a host has a restrictive execution policy set through GPO then the `win_updates` module will fail to start `task.ps1` which is run through the task scheduler. By having the new process read the file manually and invoke it using a scriptblock it can avoid this restriction and run like normal.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_updates


Omnialias:
I'm the person on Reddit that brought this issue up. I have tested this and it seems to be working. Currently running updates, which it wasn't getting to before.

jborean93:
Thanks for the confirmation, I really appreciate it.

PR #334, opened 2022-02-01
win_reboot - enforce minimum pre reboot time to 2 seconds

jborean93:
##### SUMMARY
CI is sometimes seeing a problem with the RPC server not being online when attempting a reboot. This adds a simple retry option in the reboot command if this comes up again.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_reboot

### DIFF
--- a/changelogs/fragments/win_reboot-pre-delay.yml
+++ b/changelogs/fragments/win_reboot-pre-delay.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_reboot - Always set a minimum of 2 seconds for ``pre_reboot_delay`` to ensure the plugin can read the result
--- a/plugins/action/win_reboot.py
+++ b/plugins/action/win_reboot.py
@@ -78,6 +78,11 @@ class ActionModule(ActionBase):
                 except TypeError as e:
                     raise AnsibleError("Invalid value given for '%s': %s." % (names[0], to_native(e)))
 
+                # Setting a lower value and kill PowerShell when sending the shutdown command. Just use the defaults
+                # if this is the case.
+                if names[0] == 'pre_reboot_delay' and value < 2:
+                    continue
+
                 parameters[names[0]] = value
 
         result = reboot_host(self._task.action, self._connection, **parameters)
--- a/plugins/modules/win_reboot.py
+++ b/plugins/modules/win_reboot.py
@@ -14,6 +14,7 @@ options:
   pre_reboot_delay:
     description:
     - Seconds to wait before reboot. Passed as a parameter to the reboot command.
+    - The minimum version is C(2) seconds and cannot be set lower.
     type: float
     default: 2
     aliases: [ pre_reboot_delay_sec ]


jborean93:
There's a consistent theme with this error. On 2012 the first reboot task succeeds without any problems. The 2nd task then fails with the following output

```
10:28 TASK [win_reboot : test with negative values for delays] ***********************
10:28 win_reboot: running command: (Get-CimInstance -ClassName Win32_OperatingSystem -Property LastBootUpTime).LastBootUpTime.ToFileTime()
10:29 win_reboot: command result - rc: 0, stdout: 132881657974942370, stderr: 
10:29 win_reboot: rebooting server...
10:29 win_reboot: running command: $host.UI.WriteErrorLine('start')
10:29 $ErrorActionPreference = 'Continue'
10:29 $host.UI.WriteErrorLine('eap set')
10:29 
10:29 if ($True) {
10:29     $host.UI.WriteErrorLine('b remove')
10:29     Remove-Item -LiteralPath 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Winlogon\AutoLogonChecked' -Force -ErrorAction SilentlyContinue
10:29     $host.UI.WriteErrorLine('a remove')
10:29 }
10:29 
10:29 $host.UI.WriteErrorLine('before')
10:29 $rc = Invoke-CimMethod -Query 'SELECT * FROM Win32_OperatingSystem' -MethodName Win32ShutdownTracker -Arguments @{
10:29     Timeout = [UInt32]0
10:29     Comment = 'Reboot initiated by Ansible'
10:29     ReasonCode = [UInt32]"0x80040001"
10:29     Flags = 6
10:29 }
10:29 $host.UI.WriteErrorLine("after: $($rc.ReturnValue)")
10:29 
10:29 $m = (New-Object ComponentModel.Win32Exception -ArgumentList @([int]$rc.ReturnValue)).Message
10:29 $host.UI.WriteErrorLine("gotten message: $m")
10:29 
10:29 ConvertTo-Json -Compress -InputObject @{
10:29     msg = $m
10:29     rc = $rc.ReturnValue
10:29     status = $?
10:29 }
10:29 $host.UI.WriteErrorLine('done')
10:30 win_reboot: command result - rc: 1, stdout: , stderr: start

10:30 eap set

10:30 b remove

10:30 a remove

10:30 before

10:30 after: 0

10:30 gotten message: The operation completed successfully
10:30 An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.ansible.windows.plugins.plugin_utils._reboot._ReturnResultException: win_reboot: Reboot command failed
10:30 fatal: [windows-2012]: FAILED! => {"changed": false, "elapsed": 0, "msg": "win_reboot: Reboot command failed", "rc": 1, "rebooted": false, "stderr": "start\r\neap set\r\nb remove\r\na remove\r\nbefore\r\nafter: 0\r\ngotten message: The operation completed successfully", "stdout": "", "unreachable": false}
10:30 
```

I can see that the reboot command was successful by the presence of `after: 0` with the value the return value from the CIM call. Interestingly there is no more stdout or stderr after `gotten message` and the return value is `1` for the raw command. The `win_reboot` plugin interprets this as a failure as it was unable to detect the result from the CIM call and the rc of the raw command was 1. At a guess the shutdown command is killing the process before it can write the output and potentially it's more apparent on 2012 due to the inherent slowness of the host compared to other ones. The subsequent retry of the command fails with `1717` because the host is either still going down or coming back up from the last reboot command that is still occurring.

At a guess the problem is the negative value of the `pre_reboot_delay` on the task is being converted to `0` and thus the timeout is so short it kills the reboot command process. There should be a minimum value of `pre_reboot_delay` of 5 to try and cover this scenario.

jborean93:
> At a guess the problem is the negative value of the pre_reboot_delay on the task is being converted to 0 and thus the timeout is so short it kills the reboot command process. There should be a minimum value of pre_reboot_delay of 5 to try and cover this scenario.

Looks like this was it, reran the tests 4 times and they did not fail as before.

PR #342, opened 2022-03-11
setup - Added ipv4, ipv6, mtu and speed data to ansible_interfaces

netsandbox:
##### SUMMARY
setup - Added ipv4, ipv6, mtu and speed data to ansible_interfaces

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
setup

##### ADDITIONAL INFORMATION
This add some more useful data to ansible_interfaces.

### DIFF
--- a/changelogs/fragments/setup_more_interface_data.yml
+++ b/changelogs/fragments/setup_more_interface_data.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- setup - Added ipv4, ipv6, mtu and speed data to ansible_interfaces
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -721,6 +721,15 @@ $factMeta = @(
                     elseif ($ipv6) {
                         $ipv6.Index
                     }
+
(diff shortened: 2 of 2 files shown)


jborean93 (approved):

jborean93:
I'm not a fan of how the `ansible_interfaces` value of `setup.ps1` is different from the POSIX `setup` module but that's not a problem of the code here. The changes you've given seem sane and work when testing locally so I'm happy to include them. Thanks for the changelog fragment for the changes as well.

netsandbox:
Thanks for merging.

When I stated recently working with Ansible and Windows, I was really surprised how different are the facts between Linux and Windows.
Maybe I can work on aligning the Windows facts to the Linux facts, if that would be something you also want.

jborean93:
The trouble with aligning these vars is that people rely on the existing structure. So changing things will most likely break their playbooks and there is no mechanism to warn people that certain vars may change in the future. That being said we could potentially look at doing something like `ansible_interfaces2` that mirror the existing POSIX facts and slowly encourage people to use that. Over time we could deprecate and remove the older facts once a mechanism to warn about that is in place.

PR #353, opened 2022-04-22
setup - Add system locale language

stefanDeveloper:
##### SUMMARY

Based on https://github.com/ansible/ansible/pull/32930, add the system locale language of Windows to the facts of Ansible for the Windows operating system. This helps to resolve different language-related differences under Windows. I've tried to add the system locale language with the constraints mentioned in the pull request. However, some time has passed since the last pull request, so I'm unsure if this solution is appropriated. Feel free to suggest.

##### ISSUE TYPE

Feature Pull Request

##### COMPONENT NAME

- `setup.ps1` Module

### DIFF
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -1056,7 +1056,7 @@ $factMeta = @(
     @{
         Subsets = 'locale'
         Code = {
-            $system_language = (Get-WmiObject -Class Win32_OperatingSystem).MUILanguages | Select -Index 0
+            $system_language = (Get-CimInstance -ClassName Win32_OperatingSystem).MUILanguages | Select-Object -Index 0
 
             $ansibleFacts.ansible_locale.system_language = $system_language
         }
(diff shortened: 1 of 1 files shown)


br-olf:
Can we have this please :)

devreyzoe:
I need this feature, too. Would appreciate it.

jborean93:
Sorry for the delay in reviewing this PR. Unfortunately I'm going to have to say no to this change due to 3 reasons:

* From a technical standpoint we try to avoid, as much as possible, the use of WMI/CIM classes to gather data
  * Out of the box only administrators can access WMI/CIM classes through a network logon breaking this fact for non-admin accounts.
  * CIM/WMI is also quite slow compared to using native calls through .NET, registry lookups, or even just a PInvoke definition.
* Unless it's to add parity to the existing `setup` facts module on Python we want to avoid adding anything new to this module and recommend using a separate module that a user supplies.
  * The more that is added here, the more time it will take to gather which is important for a module that runs on pretty much any connection by default
* If someone desires this information they can either put it in their own module in a collection and add it to the [FACTS_MODULES](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#facts-modules) list
  * This config allows you to configure what facts modules are run automatically in Ansible and can be extended to run your own which could include this fact
  * You can also just manually run it just like any other module

We appreciate the time and effort you've put into this and while this may not be the outcome you are looking for, I wish you the best in the future.

PR #356, opened 2022-05-10
Do not download http target before checking for state when product_id…

ptilin:
… is defined

##### SUMMARY
When win_package/product_id is defined - do not download package blob from HTTP location before making check if package is installed already.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_package.ps1

##### ADDITIONAL INFORMATION
Scope mess.


jborean93:
Looks like while this is a bug there is some code that still needs to be run if it was a URL but the package wasn't installed. The error is complaining that the path being set is not a valid Windows application as it's a URL.

PR #359, opened 2022-05-20
Fix escaping of backslash character

brianhelba:
##### SUMMARY
Currently, this backslash is not being rendered: https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_copy_module.html#parameter-dest

##### ISSUE TYPE
- Docs Pull Request


agibson2:
I am curios why this wasn't merged.  It looks like it is still not being displayed in the document referenced...

If dest is a nonexistent path, it will only be created if dest ends with “/” or “", or src is a directory.

It likely should be...

If dest is a nonexistent path, it will only be created if dest ends with “/” or “\", or src is a directory.

PR #365, opened 2022-06-07
docs win_find_module correct sample for value path

jamesnswithers:
##### SUMMARY

Corrects an error in the documentation for win_find_module where the returned value has an incorrect sample.

Current sample shows what would be the owner of the file, but should show a path to the found file.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

win_find_module


jborean93 (approved):
Thanks for the pickup here and fixing!

PR #377, opened 2022-06-16
win_acl_inheritance - added registry keys support

inorangestylee:
##### SUMMARY
Unfortunately, this module does not support changing the inheritance of registry keys. There are several reasons for this:

- use of the `path` type for the `path` parameter, which does not work with registry keys. (Has been replaced by `str`)
- use of `LiteralPath` parameter which does not work correctly with registry paths in Powershell 5.1 and was fixed in https://github.com/PowerShell/PowerShell/issues/11566 (was replaced by `Path`).

Changes here:
- registry path support added
- refactored from legacy to modern module style
- 'try-catch' block redesigned to handle errors of specific calls
- redesigned inheritance and explicit rules equality check

*No breaking changes. No need to rewrite playbooks*

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_acl_inheritance

### DIFF
--- a/plugins/modules/win_acl_inheritance.ps1
+++ b/plugins/modules/win_acl_inheritance.ps1
@@ -1,75 +1,156 @@
 #!powershell
 
+# Copyright: (c) 2022, Oleg Galushko (@inorangestylee)
 # Copyright: (c) 2015, Hans-Joachim Kliemeck <git@kliemeck.de>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-#Requires -Module Ansible.ModuleUtils.Legacy
+#AnsibleRequires -CSharpUtil Ansible.Basic
 
-$params = Parse-Args $args -supports_check_mode $true
-$check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -default $false
-
-$result = @{
-    changed = $false
+$spec = @{
+    options = @{
+        path = @{ type = 'str'; required = $true }
+        reorganize = @{ type = 'bool'; default = $false }
+        state = @{ type = 'str'; default = 'absent'; choices = @('absent', 'present') }
+    }
+    supports_check_mode = $true
 }
 
-$path = Get-AnsibleParam -obj $params "path" -type "path" -failifempty $true
-$state = Get-AnsibleParam -obj $params "state" -type "str" -default "absent" -validateSet "present", "absent" -resultobj $result
-$reorganize = Get-AnsibleParam -obj $params "reorganize" -type "bool" -default $false -resultobj $result
+function Get-AclEx {
+    [CmdletBinding()]
+    [OutputType([System.Security.AccessControl.NativeObjectSecurity])]
+    param(
+        [System.String] $LiteralPath
+    )
 
-If (-Not (Test-Path -LiteralPath $path)) {
-    Fail-Json $result "$path file or directory does not exist on the host"
+    $item = Get-Item -LiteralPath $LiteralPath
+    return $item.GetAccessControl()
 }
 
-Try {
-    $objACL = Get-ACL -LiteralPath $path
-    # AreAccessRulesProtected - $false if inheritance is set ,$true if inheritance is not set
-    $inheritanceDisabled = $objACL.AreAccessRulesProtected
+function Set-AclEx {
+    [CmdletBinding(SupportsShouldProcess)]
+    param(
+        [System.String] $LiteralPath,
+        [System.Security.AccessControl.NativeObjectSecurity] $AclObject
+    )
 
-    If (($state -eq "present") -And $inheritanceDisabled) {
-        # second parameter is ignored if first=$False
-        $objACL.SetAccessRuleProtection($False, $False)
+    $item = Get-Item -LiteralPath $LiteralPath
+    if ($PSCmdlet.ShouldProcess($LiteralPath)) {
+        if ($item.PSProvider.Name -eq 'Registry') {
+            Set-Acl -LiteralPath $LiteralPath -AclObject $AclObject
+        }
+        else {
+            $item.SetAccessControl($AclObject)
+        }
+    }
+    else {
+        Write-Verbose "WhatIf: Performing setting ACL on '$literalPath'"
+    }
+}
 
-        If ($reorganize) {
-            # it wont work without intermediate save, state would be the same
-            Set-ACL -LiteralPath $path -AclObject $objACL -WhatIf:$check_mode
-            $result.changed = $true
-            $objACL = Get-ACL -LiteralPath $path
+function Remove-AclExplicitDuplicate {
+    [CmdletBinding(SupportsShouldProcess)]
+    [OutputType([System.Security.AccessControl.NativeObjectSecurity])]
+    param (
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+        [System.Security.AccessControl.NativeObjectSecurity] $acl
+    )
 
-            # convert explicit ACE to inherited ACE
-            ForEach ($inheritedRule in $objACL.Access) {
-                If (-not $inheritedRule.IsInherited) {
-                    Continue
-                }
+    Begin {}
 
-                ForEach ($explicitRrule in $objACL.Access) {
-                    If ($explicitRrule.IsInherited) {
-                        Continue
+    Process {
+        $properties = $acl.Access |
+            Get-Member -MemberType Property |
+            Where-Object { $_.Name -ne 'IsInherited' } |
+            Select-Object -ExpandProperty Name
+
+        ForEach ($inheritedRule in $($acl.Access | Where-Object { $_.IsInherited })) {
+            ForEach ($explicitRule in $($acl.Access | Where-Object { -not $_.IsInherited })) {
+                If ($null -eq (Compare-Object -ReferenceObject $explicitRule -DifferenceObject $inheritedRule -Property $properties)) {
+                    if ($PSCmdlet.ShouldProcess($acl)) {
+                        $acl.RemoveAccessRule($explicitRule)
                     }
-
-                    If (
-                        ($inheritedRule.FileSystemRights -eq $explicitRrule.FileSystemRights) -And
-                        ($inheritedRule.AccessControlType -eq $explicitRrule.AccessControlType) -And
-                        ($inheritedRule.IdentityReference -eq $explicitRrule.IdentityReference) -And
-                        ($inheritedRule.InheritanceFlags -eq $explicitRrule.InheritanceFlags) -And
-                        ($inheritedRule.PropagationFlags -eq $explicitRrule.PropagationFlags)
-                    ) {
-                        $objACL.RemoveAccessRule($explicitRrule)
+                    else {
+                        Write-Verbose "WhatIf: Performing remove explicit duplicate rule: $explicitRule"
                     }
                 }
             }
         }
+        return $acl
+    }
 
-        Set-ACL -LiteralPath $path -AclObject $objACL -WhatIf:$check_mode
-        $result.changed = $true
-    }
-    Elseif (($state -eq "absent") -And (-not $inheritanceDisabled)) {
-        $objACL.SetAccessRuleProtection($True, $reorganize)
-        Set-ACL -LiteralPath $path -AclObject $objACL -WhatIf:$check_mode
-        $result.changed = $true
-    }
-}
-Catch {
-    Fail-Json $result "an error occurred when attempting to disable inheritance: $($_.Exception.Message)"
+    End {}
 }
 
-Exit-Json $result
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$path = $module.Params.path
+$reorganize = $module.Params.reorganize
+$state = $module.Params.state
+$check_mode = $module.CheckMode
+
+$module.Result.changed = $false
+
+$pathQualifier = Split-Path -Path $path -Qualifier -ErrorAction SilentlyContinue
+
+if ($pathQualifier -eq 'HKCR:' -and (-not (Test-Path -LiteralPath HKCR:\))) {
+    $null = New-PSDrive -Name HKCR -PSProvider Registry -Root HKEY_CLASSES_ROOT
+}
+if ($pathQualifier -eq 'HKU:' -and (-not (Test-Path -LiteralPath HKU:\))) {
+    $null = New-PSDrive -Name HKU -PSProvider Registry -Root HKEY_USERS
+}
+if ($pathQualifier -eq 'HKCC:' -and (-not (Test-Path -LiteralPath HKCC:\))) {
+    $null = New-PSDrive -Name HKCC -PSProvider Registry -Root HKEY_CURRENT_CONFIG
+}
+
+If (-Not (Test-Path -LiteralPath $path)) {
+    $module.FailJson("$path does not exist")
+}
+
+try {
+    $acl = Get-AclEx -LiteralPath $path
+}
+catch {
+    $module.FailJson($_.ToString())
+}
+
+$module.Diff.before = @{ access_rules_protected = $acl.AreAccessRulesProtected }
+
+If (($state -eq 'present') -and $acl.AreAccessRulesProtected) {
+    try {
+        $acl.SetAccessRuleProtection($false, $false)
+    }
+    catch {
+        $module.FailJson($_.ToString())
+    }
+    if ($reorganize) {
+        Set-AclEx -LiteralPath $path -AclObject $acl -WhatIf:$check_mode
+        $acl = Remove-AclExplicitDuplicate($acl)
+    }
+    Set-AclEx -LiteralPath $path -AclObject $acl -WhatIf:$check_mode
+    $module.Result.changed = $true
+}
+
+if (($state -eq 'absent') -and (-not $acl.AreAccessRulesProtected)) {
+    try {
+        $acl.SetAccessRuleProtection($true, $reorganize)
+    }
+    catch {
+        $module.FailJson($_.ToString())
+    }
+    Set-AclEx -LiteralPath $path -AclObject $acl -WhatIf:$check_mode
+    $module.Result.changed = $true
+}
+
+if (-not $check_mode) {
+    try {
+        $acl = Get-AclEx -LiteralPath $path
+    }
+    catch {
+        $module.FailJson($_.ToString())
+    }
+}
+
+$module.Diff.after = @{ access_rules_protected = $acl.AreAccessRulesProtected }
+$module.Result.access_rules_protected = $acl.AreAccessRulesProtected
+
+$module.ExitJson()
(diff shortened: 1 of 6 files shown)


plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
+
+If (-Not (Test-Path -LiteralPath $path)) {
+    $module.FailJson("$path does not exist")
+}
+
+try {
+    $acl = Get-AclEx -LiteralPath $path
+}
+catch {
+    $module.FailJson($_.ToString())
```
jborean93, line 113: It's best to provide a message saying what this was trying to do with the exception message at the end. The `FailJson` also accepts an optional 2nd argument which is the Exception or ErrorRecord which is used to display the full error + stacktrace when running with a higher verbosity.

```suggestion
    $module.FailJson("Failed to find '$path': $($_.ToString())", $_)
```

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
+}
+
+$module.Diff.before = @{ access_rules_protected = $acl.AreAccessRulesProtected }
```
jborean93, line 116: Typically we try and make the diff follow the same keys as the module options. In this case it would be `@{ state = 'present or absent' }`

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
+If (($state -eq 'present') -and $acl.AreAccessRulesProtected) {
+    try {
+        $acl.SetAccessRuleProtection($false, $false)
     }
-    Elseif (($state -eq "absent") -And (-not $inheritanceDisabled)) {
-        $objACL.SetAccessRuleProtection($True, $reorganize)
-        Set-ACL -LiteralPath $path -AclObject $objACL -WhatIf:$check_mode
-        $result.changed = $true
+    catch {
+        $module.FailJson($_.ToString())
```
jborean93, line 123: Same thing here, include a brief error message that describes the action and also pass in $_ as the 2nd argument.

You can also just omit the whole try/catch block. An uncaught exception will be displayed with the exception details on higher verbosities as well.

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
 }
-Catch {
-    Fail-Json $result "an error occurred when attempting to disable inheritance: $($_.Exception.Message)"
+
+if (($state -eq 'absent') -and (-not $acl.AreAccessRulesProtected)) {
+    try {
+        $acl.SetAccessRuleProtection($true, $reorganize)
+    }
+    catch {
+        $module.FailJson($_.ToString())
```
jborean93, line 138: Same thing here, include a brief error message that describes the action and also pass in $_ as the 2nd argument.

You can also just omit the whole try/catch block. An uncaught exception will be displayed with the exception details on higher verbosities as well.

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
+    $module.Result.changed = $true
 }
 
-Exit-Json $result
+if (-not $check_mode) {
+    try {
+        $acl = Get-AclEx -LiteralPath $path
+    }
+    catch {
+        $module.FailJson($_.ToString())
```
jborean93, line 149: Same thing here, include a brief error message that describes the action and also pass in $_ as the 2nd argument.

You can also just omit the whole try/catch block. An uncaught exception will be displayed with the exception details on higher verbosities as well.

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
+    }
+}
+
+$module.Diff.after = @{ access_rules_protected = $acl.AreAccessRulesProtected }
+$module.Result.access_rules_protected = $acl.AreAccessRulesProtected
```
jborean93, line 154: Is there any benefit to returning this. The value seems to just be whatever `state` was equal to. Typically for modules we try and only return the information the caller doesn't know about, like metadata on the item being manipulated here.

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
-    Elseif (($state -eq "absent") -And (-not $inheritanceDisabled)) {
-        $objACL.SetAccessRuleProtection($True, $reorganize)
-        Set-ACL -LiteralPath $path -AclObject $objACL -WhatIf:$check_mode
-        $result.changed = $true
+    catch {
+        $module.FailJson($_.ToString())
     }
+    if ($reorganize) {
+        Set-AclEx -LiteralPath $path -AclObject $acl -WhatIf:$check_mode
+        $acl = Remove-AclExplicitDuplicate($acl)
```
jborean93, line 127: Try and keep the `-parameter value` syntax, this avoids confusion with invoking .NET functions. I believe you also want `-WhatIf:$check_mode` as well here.

plugins/modules/win_acl_inheritance.py
```diff
@@ -14,8 +15,8 @@
   path:
     description:
       - Path to be used for changing inheritance
```
jborean93, line 17: Should add a new entry saying

> Support for registry keys have been added in C(ansible.windows>=1.11.0).

jborean93 (commented):
Sorry for the delay in looking at the PR. The changes so far look good but I've added a few questions and suggestions. The PR also needs a changelog fragment to document the new changes being made (registry key support) and having integration tests are also required. You can find the existing integration tests at https://github.com/ansible-collections/ansible.windows/tree/main/tests/integration/targets/win_acl_inheritance

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,156 @@
...
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-#Requires -Module Ansible.ModuleUtils.Legacy
+#AnsibleRequires -CSharpUtil Ansible.Basic
 
-$params = Parse-Args $args -supports_check_mode $true
-$check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -default $false
+$spec = @{
+    options = @{
+        path = @{ type = 'str'; required = $true }
```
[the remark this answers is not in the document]
briantist: How does the `path` (ansible) type prevent it from working with registry keys? It still returns a string, with the vars already expanded, so I can't see why it would fail. Being able to expand env vars could in some circumstances be useful even with reg paths, but more importantly, it ensures that the the documentation is still clear.

If this does need to change type to `str`, let's also ensure the option's description notes that it still expands paths.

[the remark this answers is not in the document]
inorangestylee: it`s throw an exception at this line:
https://github.com/ansible/ansible/blob/e4890afc4e2af6c5126d056a2b4fdb1f1f122cf6/lib/ansible/module_utils/csharp/Ansible.Basic.cs#L594
```
Path.GetFullPath(stringValue);
```
Here is different behavior in PS5 and PS7. An exception is thrown if there is a ":" inside the string
```powershell
PS C:\Users\orange> $PsVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.22000.832
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.22000.832
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKEY_LOCAL_MACHINE\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
Exception calling "GetFullPath" with "1" argument(s): "The given path's format is not supported."
At line:1 char:1
+ [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE:\SOFTWARE\AdminAbii ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : NotSupportedException

PS C:\Users\orange>
PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
Exception calling "GetFullPath" with "1" argument(s): "The given path's format is not supported."
At line:1 char:1
+ [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE:\SOFTWARE\AdminAbii ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : NotSupportedException

PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKEY_LOCAL_MACHINE\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKLM:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
Exception calling "GetFullPath" with "1" argument(s): "The given path's format is not supported."
At line:1 char:1
+ [System.IO.Path]::GetFullPath("HKLM:\SOFTWARE\AdminAbiit\win_acl_inhe ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : NotSupportedException

PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKLM\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKLM\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
```
and 
```powershell
PS C:\Users\orange> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.2.5
PSEdition                      Core
GitCommitId                    7.2.5
OS                             Microsoft Windows 10.0.22000
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKEY_LOCAL_MACHINE\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKEY_LOCAL_MACHINE:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKEY_LOCAL_MACHINE:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKLM:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKLM:\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
PS C:\Users\orange> [System.IO.Path]::GetFullPath("HKLM\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)")
C:\Users\orange\HKLM\SOFTWARE\AdminAbiit\win_acl_inheritance .ÅÑŚÌβŁÈ [$!@^&test(;)
```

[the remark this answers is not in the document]
jborean93: Yea sorry for not replying, if you have the type as `path` it will attempt to validate it as an actual file path which fails on registry paths (more than 1 char for the root).

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,158 @@
...
 
-If (-Not (Test-Path -LiteralPath $path)) {
-    Fail-Json $result "$path file or directory does not exist on the host"
+    $item = Get-Item -LiteralPath $LiteralPath
+    if ($PSCmdlet.ShouldProcess($LiteralPath)) {
+        if ($item.PSProvider.Name -eq 'Registry') {
+            Set-Acl -LiteralPath $LiteralPath -AclObject $AclObject
+        }
+        else {
+            $item.SetAccessControl($AclObject)
```
jborean93, line 42: I'm curious why `SetAccessControl` can't be used on a registry key? It seems like the type has the method defined https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.setaccesscontrol?view=net-6.0. Does `Set-Acl` do more work here?

plugins/modules/win_acl_inheritance.ps1
```diff
@@ -1,75 +1,158 @@
...
+}
 
-        Set-ACL -LiteralPath $path -AclObject $objACL -WhatIf:$check_mode
-        $result.changed = $true
+$pathQualifier = Split-Path -Path $path -Qualifier -ErrorAction SilentlyContinue
+$pathQualifier = $pathQualifier.Replace(':', '')
+
+if ($pathQualifier -in $regeditHives.Keys -and (-not (Test-Path -LiteralPath "${pathQualifier}:\"))) {
+    $null = New-PSDrive -Name $pathQualifier -PSProvider 'Registry' -Root $regeditHives.$pathQualifier
+    Push-Location -LiteralPath "${pathQualifier}:\"
```
jborean93, line 102: I seem to recall this is needed to do some `Set-Acl` on registry paths, is that actually true. While we do run code in Runspaces which shouldn't affect anything else I'm not the biggest fan of this. If we can use `$item.SetAccessControl($acl)` to set the info is this still needed?

If so you could probably do

```powershell
New-PSDrive -Name $pathQualifier -PSProvider 'Registry' -Root $regeditHives.$pathQualifier | Push-Location
```

tests/integration/targets/win_acl_inheritance/tasks/main.yml
```diff
@@ -2,14 +2,13 @@
 # Test setup
 # Use single task to save in CI runtime
 - name: create test folders
-  win_shell: |
+  ansible.windows.win_shell: |
```
jborean93, line 5: Just an FYI you don't need to use the fully qualified name for things in this collection. You don't need to change it back but I'm unsure if you did this manually for some reason or it was changed due to another linting tool.

jborean93 (commented):
The changes look good, appreciate the time and effort you've put into this and adding the tests. I've only got a few tiny questions, the only remaining thing left would be to create a changelog fragement with something like (feel free to adjust how you wish)

```yaml
minor_changes:
- win_acl_inheritance - support setting inheritance rules on registry keys
```

plugins/modules/win_acl_inheritance.ps1
inorangestylee: I don't know if this is a bug or a feature, but the problem has long been mentioned on various boards. Trying to apply ACL leads to an exception. Most likely the problem is a lack of `SE*` privileges (my guess, as I could not get below the identifier of `Microsoft.Win32.RegistryKey public RegistryKey? OpenSubKey(string name, RegistryRights rights`). `Get-Item` opens the key without `WriteAccess` rights. I've done some research with reflexion and can probably confirm this.

PS7: no WriteAccess flag
```Powershell
$path = "HKLM:\SOFTWARE\Test"
$key = get-item -LiteralPath $path
$state = $key.GetType().GetField("_state", [System.Reflection.BindingFlags]::Instance -bor [System.Reflection.BindingFlags]::NonPublic)
$state.GetValue($key)

0
```

PS5: no WriteAcces flag
```Powershell
$path =  "HKLM:\SOFTWARE\Test"
$key = get-item -LiteralPath $path
$state = $key.GetType().GetField("state", [System.Reflection.BindingFlags]::Instance -bor [System.Reflection.BindingFlags]::NonPublic)
$state.GetValue($key)

0
```

.Net native
```Powershell
# w/o write access flag (default)
$key = [Microsoft.Win32.Registry]::LocalMachine.OpenSubKey('SOFTWARE\Test')
$state = $key.GetType().GetField("_state", [System.Reflection.BindingFlags]::Instance -bor [System.Reflection.BindingFlags]::NonPublic)
$state.GetValue($key)

0

# w/ write access flag
$key = [Microsoft.Win32.Registry]::LocalMachine.OpenSubKey('SOFTWARE\Test', $true)
$state = $key.GetType().GetField("_state", [System.Reflection.BindingFlags]::Instance -bor [System.Reflection.BindingFlags]::NonPublic)
$state.GetValue($key)

WriteAccess
```

tests/integration/targets/win_acl_inheritance/tasks/main.yml
inorangestylee: Just a habit :) After the linter recommendations, it became a habit for me, I use them in all my roles and playbooks.

jborean93: Fair enough, I assumed it was something like that.

plugins/modules/win_acl_inheritance.ps1
jborean93: Ah ok that makes sense that `Get-Item` won't open it with the access required. Just an FYI we do have some shared utils https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/powershell/Ansible.ModuleUtils.PrivilegeUtil.psm1 that can be used to enable privileges on a process as needed.

jborean93 (approved):

jborean93:
Thanks for working on this.

Issue #417, opened 2022-09-26
win_updates fails because temporary file does not exist


anasaizg:
##### SUMMARY

In some servers, win_updates fails with the following message: "Cannot find path 'C:\\Users\\TEMP.<User>\\AppData\\Local\\Temp\\ansible-tmp-*' because it does not exist.". However, when update is launched manually, there seems to be no errors.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_updates

##### ANSIBLE VERSION

```
ansible 2.10.14
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/my-envs/venv-windows/lib64/python3.6/site-packages/ansible
  executable location = /opt/my-envs/venv-windows/bin/ansible
  python version = 3.6.8 (default, Nov 18 2021, 10:07:16) [GCC 4.8.5 20150623 (Red Hat 4.8.5-44.0.3)]

```

##### COLLECTION VERSION

```
Collection      Version
--------------- -------
ansible.windows 1.11.1

```

##### OS / ENVIRONMENT

Target OS: Microsoft Windows Server 2016 Standard
Powershell version: 5
Connection Type: PSRP

##### STEPS TO REPRODUCE

```
- name: Windows updates - log to {{ansible_log}}
  ansible.windows.win_updates:
    category_names: "{{ category_updates |default('*') }}"
    state: installed
    log_path: "{{ ansible_log }}"
    reboot: yes
    reboot_timeout: 3600
  register: win_updates_result
  ignore_errors: yes

- debug:
    var: win_updates_result
```

##### EXPECTED RESULTS
Server is updated after some minutes (or hours, depending on the patches)

##### ACTUAL RESULTS
Update task fails with the following output:

```
{
[...]

PR #422, opened 2022-10-04
win_updates - handle deleted temp profile

jborean93:
##### SUMMARY
When the windows logon has a temp profile used, the path is deleted when a reboot occurs which causes an error as the next run tries to use the same `tmpdir` created at the beginning. This adds an explicit error flag that the action plugin can use to recreate a new tmpdir used by the module if the path does not exist.

Fixes: https://github.com/ansible-collections/ansible.windows/issues/417

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_updates


plugins/action/win_updates.py
```diff
@@ -831,7 +835,17 @@ def _run_sync(self, task_vars, module_options, reboot, reboot_timeout):  # type:
...
-            update_result = self._run_updates(task_vars, module_options, poll_script_path, cancel_script_path)
+            try:
+                update_result = self._run_updates(task_vars, module_options, poll_script_path, cancel_script_path)
+            except _RecreateTempPathException:
+                display.vv("Failure when running win_updates module with existing tempdir, retrying with new dir")
+                self._connection._shell.tmpdir = None
+                self._make_tmp_path()
+                module_options['_output_path'] = self._connection._shell.tmpdir
+                poll_script_path = self._copy_script(_POLL_SCRIPT, 'poll.ps1')
+                cancel_script_path = self._copy_script(_CANCEL_SCRIPT, 'cancel.ps1')
```
nitzmahone, line 846: Only nit here is it feels wrong to duplicate the creation vs just making it a function, otherwise seems pretty reasonable

jborean93: True, it does mean I have to return 3 values but it makes sense to try and collect all the operations for setting up a tempdir for use in the module in one location.

PR #423, opened 2022-10-07
Add support to apply ACL to Junctions 

gillg:
##### SUMMARY
Currently applying ACLs on a SymLink or Junction just applies the ACLs to the "link" not to the target.
There was no way to follow links, what seems to be the first intent when you set ACLs

I didn't added a new official variable for now, I let you decide if it's necessary or not.

### DIFF
--- a/plugins/modules/win_acl.ps1
+++ b/plugins/modules/win_acl.ps1
@@ -8,6 +8,7 @@
 #Requires -Module Ansible.ModuleUtils.Legacy
 #Requires -Module Ansible.ModuleUtils.PrivilegeUtil
 #Requires -Module Ansible.ModuleUtils.SID
+#Requires -Module Ansible.ModuleUtils.LinkUtil
 
 $ErrorActionPreference = "Stop"
 
@@ -89,7 +90,7 @@ $state = Get-AnsibleParam -obj $params -name "state" -type "str" -default "prese
 
 $inherit = Get-AnsibleParam -obj $params -name "inherit" -type "str"
 $propagation = Get-AnsibleParam -obj $params -name "propagation" -type "str" -default "None" -validateset "InheritOnly", "None", "NoPropagateInherit"
-$follow_links = $True
+$follow = Get-AnsibleParam -obj $params -name "follow" -type "bool" -default "true"
 
 # We mount the HKCR, HKU, and HKCC registry hives so PS can access them.
 # Network paths have no qualifiers so we use -EA SilentlyContinue to ignore that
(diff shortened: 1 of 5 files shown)


plugins/modules/win_acl.ps1
```diff
@@ -103,6 +104,14 @@ if ($path_qualifier -eq "HKCC:" -and (-not (Test-Path -LiteralPath HKCC:\))) {
     New-PSDrive -Name HKCC -PSProvider Registry -Root HKEY_CURRENT_CONFIG > $null
 }
 
+$path = Resolve-Path -LiteralPath $path
+Do {
+    $PathDetails = Get-ItemProperty -LiteralPath $path
+    if ($PathDetails.Target -and $PathDetails.Target.GetType().Name -eq "String[]" -and $follow_links) {
```
jborean93, line 110: I believe `Target` is a property added in WinPS 5.1 but we still need to support pwsh 3 and 4. We have some module utils that can help with links.

```powershell
# Put at the top of the script with #AnsibleRequires to reference this module util
#Requires -Module Ansible.ModuleUtils.LinkUtil

...

$follow_links = $true

while ($follow_links) {
    try {
        $link_info = Get-Link -Path $path
    } catch {
        $link_info = $null
    }

    if ($link_info -and $link_info.Type -in @("SymbolicLink", "JunctionPoint")) {
        $path = $link_info.AbsolutePath
    }
    else {
        break
    }
}
```

plugins/modules/win_acl.ps1
```diff
@@ -89,6 +89,7 @@ $state = Get-AnsibleParam -obj $params -name "state" -type "str" -default "prese
 
 $inherit = Get-AnsibleParam -obj $params -name "inherit" -type "str"
 $propagation = Get-AnsibleParam -obj $params -name "propagation" -type "str" -default "None" -validateset "InheritOnly", "None", "NoPropagateInherit"
+$follow_links = $True
```
jborean93, line 92: This should be a parameter, we typically use `follow` for this type of option as per https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-follow. Don't forget to update the `win_acl.py` docs for this new parameter.

jborean93 (commented):
It would be great if we could have some tests to cover this scenario and a changelog as per https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to.

plugins/modules/win_acl.ps1
gillg: Sorry but I don't understand the logic of your loop.
You never update $follow_links inside the loop so it should be an infinite loop no ?
My initial idea was to walk over symlinks/junctions.

Let's imagine a Symlink pointing on a Junction wich is a valid use case.

jborean93: The `while ($follow_links)` will ensure the code is run if `$follow_links` is `$true`. It will continuously loop through the code until either `Get-Link` fails or there is no link target. This is done through the if check at the end where `break` is called to exit the loop if either `$link_info` is not set (catch block or the path is not a link) or the link type is not a `SymbolicLink` or `JunctionPoint`. This ensures that it will continue to follow the link's targets until that condition is met.

jborean93:
Sorry I led you a tiny bit astray in my recommended example. When writing some tests for this functionality I realised a few things were missing. I've since pushed a commit that:

* Fixed my incorrect suggestion
* Added tests
* Changed the default for `follow` to `false`

Unfortunately having `follow: true` as a default would be a breaking change for this module so I've had to change the default to `$false`. This keeps things aligned with the `win_stat`, `stat` modules as well.

Thanks for working on this one though!

gillg:
Thank you a lot @jborean93 I would not be comfortable to write the tests anyway!
And fixing my English in the changelog is always a good idea 😁

PR #424, opened 2022-10-07
Allows to force reimport a cert if you added a private key with win_certificate_store

gillg:
##### SUMMARY
If you try to import a cert in the cert store wich already exists but without a private key associated, you can't because Ansible will detect no change at thumbprint level.
With that change ansible will force reimport if a private key is present during the second import.

We could imagine also force the import, if at the opposite, we remove the private key later. To be discussed.

### DIFF
--- a/plugins/modules/win_certificate_store.ps1
+++ b/plugins/modules/win_certificate_store.ps1
@@ -479,7 +479,17 @@ try {
         foreach ($cert in $certs) {
             $module.Result.thumbprints += $cert.Thumbprint
             $found_certs = $store_certificates.Find([System.Security.Cryptography.X509Certificates.X509FindType]::FindByThumbprint, $cert.Thumbprint, $false)
-            if ($found_certs.Count -eq 0) {
+            $force_reimport = $False
+            if ($found_certs.Count -ne 0 -and $cert.HasPrivateKey) {
+                # Because the thumbprint doesn't identify a private key, if the cert to import contains a PrivateKey
+                # we have to update the existing cert, reimporting the cert + key.
+                foreach ($found_cert in $found_certs) {
+                    if ($cert.HasPrivateKey -and -not $found_cert.HasPrivateKey) {
+                        $force_reimport = $True
+                    }
+                }
+            }
+            if ($found_certs.Count -eq 0 -or $force_reimport) {
                 try {
                     if (-not $module.CheckMode) {
                         $store.Add($cert)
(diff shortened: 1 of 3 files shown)


plugins/modules/win_certificate_store.ps1
```diff
@@ -479,7 +479,17 @@ try {
...
             $module.Result.thumbprints += $cert.Thumbprint
             $found_certs = $store_certificates.Find([System.Security.Cryptography.X509Certificates.X509FindType]::FindByThumbprint, $cert.Thumbprint, $false)
-            if ($found_certs.Count -eq 0) {
+            $force_reimport = $False
+            if ($found_certs.Count -ne 0 -and $cert.HasPrivateKey) {
+                # Because the thumbprint doesn't identify a private key, if the cert to import contains a PrivateKey
+                # we have to update the existing cert, reimporting the cert + key.
+                foreach ($found_cert in $found_certs) {
+                    if ($cert.HasPrivateKey -and -not $found_cert.HasPrivateKey) {
+                        $force_reimport = $True
```
jborean93, line 488: Is there a way to compare the 2 private keys rather than always import a cert with a private key? I'm just thinking of the idempotency scenarios here.

jborean93 (changes requested):
Sorry for the delay in the review here. I've been travelling for AnsibleFest and recently had to take some time off after.

plugins/modules/win_certificate_store.ps1
gillg: In my opinion it makes no sense by essence.
If you have a different private key, you can't generate the same certificate, so not the same thumbprint too.
Moreover very often by default the private keys are "sealed" so we can't read them in powershell.

jborean93: Ah ok, I think I understand the case here. Essentially a cert has been imported but without the private key but this scenario is when you try to import the cert again but with the key present. Before Ansible was only checking based on the thumbprint, now you do the thumbprint check but also whether a key is present on the imported cert.

Will `$found_cert.HasPrivateKey` be true if there is a private key present but just not marked as exportable?

jborean93: > Will $found_cert.HasPrivateKey be true if there is a private key present but just not marked as exportable?

Just tested it locally and yes a non-exportable private key has `HasPrivateKey` still set to `$true`.

jborean93 (approved):

PR #455, opened 2022-12-01
Clarify host_out and output differences

nicozanf:
I could not find these useful information in other places ;-)

##### ISSUE TYPE
- Docs Pull Request

+label: docsite_pr

##### SUMMARY

It's quite self-explaining ....

### DIFF
--- a/plugins/modules/win_powershell.py
+++ b/plugins/modules/win_powershell.py
@@ -102,8 +102,8 @@ notes:
   based on verbosity. Added in C(1.9.0).
 - Any host/console direct output like C(Write-Host) or C([Console]::WriteLine) is not considered an output object, they are
   returned as a string in I(host_out) and I(host_err).
-- Any host/console pipelined output like C(Write-Output) or C(echo) is also not considered an output object, but they are
-  returned as a list in I(host_output).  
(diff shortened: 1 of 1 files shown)


plugins/modules/win_powershell.py
```diff
@@ -100,8 +100,10 @@
   module has finished.
 - C($Ansible.Verbosity) reveals Ansible's verbosity level for this play. Allows the script to set VerbosePreference/DebugPreference
   based on verbosity. Added in C(1.9.0).
-- Any host/console output like C(Write-Host) or C([Console]::WriteLine) is not considered an output object, they are
+- Any host/console direct output like C(Write-Host) or C([Console]::WriteLine) is not considered an output object, they are
   returned as a string in I(host_out) and I(host_err).
+- Any host/console output stream like C(Write-Output) or C(echo) is also not considered an output object, but they are
```
jborean93, line 105: This isn't really true, by definition `Write-Output` is output that is sent on the output stream. The `echo` command is just an alias for `Write-Output`. I'm happy to keep the explicit mention that `Write-Output/echo` is returned under `output` but would need to see:

* The part about it not being considered an output object removed
* A mention that implicit output is also returned, i.e. `Write-Output "foo"` and `"foo"` are the same.

PR #474, opened 2023-03-03
win_package more examples for different scenarios

Jvekka:
##### SUMMARY
Two examples added to win_package to provide better examples for end users which doesn't read and understand the docs so well (like me).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
win_package

##### ADDITIONAL INFORMATION
Added couple examples which were asked (multiple) times in Reddit and Stack Overflow.
- MSSQL installation to provide example for DPAPI
- Application example how to handle multiple properties for installer

```

### DIFF
--- a/plugins/modules/win_package.py
+++ b/plugins/modules/win_package.py
@@ -274,7 +274,6 @@ EXAMPLES = r'''
       - SAPWD=VeryHardPassword
       - /ConfigurationFile=C:\temp\configuration.ini
   become: yes
-  become_flags: logon_type=batch
   vars:
     ansible_become_method: runas
     ansible_become_user: "{{ user }}"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: plugins/modules/win_package.py [hunk 1]
```diff
@@ -256,6 +256,34 @@
...
 
+- name: Install Application from msi with multiple properties for installer
+  ansible.windows.win_package:
+    path: C:\temp\Application.msi
+    state: present
+    arguments: >-
+      /quiet
+      /norestart
+      /log
+      C:\application_server.log
```
jborean93, line 267:
```suggestion
```

These 4 arguments are already specified when using an MSI in `win_package` and should not be explicitly set.

plugins/modules/win_package.py
```diff
@@ -256,6 +256,34 @@
...
+
+- name: Install Microsoft® SQL Server® 2019 Express (DPAPI example)
+  ansible.windows.win_package:
+    path: C:\temp\SQLEXPR_x64_ENU\SETUP.EXE
+    product_id: Microsoft SQL Server SQL2019
+    arguments:
+      - SAPWD=VeryHardPassword
+      - /ConfigurationFile=C:\temp\configuration.ini
+  become: yes
+  become_flags: logon_type=batch
```
jborean93, line 281: Any reason for this particular logon flag?

Jvekka: Took some time to verify. It did work actually without the flags as well.

jborean93 (approved):

PR #478, opened 2023-03-10
win updates include is downloaded property and summary property for all updates

MartinSiddall:
##### SUMMARY

When using a managed server, approvals are done centrally and the client machines poll the SUS server on a regular basis. If updates are available our GPO settings tell the client machine to download the updates ready for installing.

As part of our WSUS automation processes we would like to determine which patches have been downloaded and as a summary if all updates have been downloaded. This will help us decide what step to process next in the automation.

To facilitate this we have included on a per update basis the is_downloaded property  and on the task itself a flag to determine if all updates have been downloaded.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_updates module py results processing

### DIFF
--- a/plugins/action/win_template.py
+++ b/plugins/action/win_template.py
@@ -7,7 +7,6 @@ __metaclass__ = type
 
 import os
 import shutil
-import stat
 import tempfile
 
 from ansible import constants as C
--- a/plugins/action/win_updates.py
+++ b/plugins/action/win_updates.py
@@ -4,7 +4,7 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import json
+import json # pylint: disable=unused-import
 import os.path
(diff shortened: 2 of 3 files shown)


jborean93:
I'm reluctant to adding more return values to this module as it's already complicated. Is this something you could instead achieve with a custom filter plugin to fit your needs. The `is_downloaded` also just seems to be a duplicate of the existing `downloaded` key that is already there. You could even use the builtin filters to get this value:

```yaml
- debug:
    msg: '{{ update_output.updates.values() | selectattr("downloaded", "equalto", false) | length | int | bool}}'
```

Essentially it will be `true` if all of the `downloaded` values of each update is `true`. If one of them is `false` then the output value is also `false.`

jborean93:
Closing as per the above message.

PR #487, opened 2023-03-31
win_file: Add support for hard link, symbolic link and junction

silid:
##### SUMMARY
Adds support to win_file to create and update hard links, symbolic links, and junctions.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_file

##### ADDITIONAL INFORMATION
- This will create or update links like-for-like but not replace other types of files.
- Symbolic links support UNC paths, so there is a regex to convert paths that start \\ to UNC\ as returned by Get-Item when comparing existing values.
- This includes some refactoring to make it easier to read and simplify the update of links.

```paste below

```

### DIFF
--- a/plugins/modules/win_file.ps1
+++ b/plugins/modules/win_file.ps1
@@ -14,7 +14,8 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -default
 $_remote_tmp = Get-AnsibleParam $params "_ansible_remote_tmp" -type "path" -default $env:TMP
 
 $path = Get-AnsibleParam -obj $params -name "path" -type "path" -failifempty $true -aliases "dest", "name"
-$state = Get-AnsibleParam -obj $params -name "state" -type "str" -validateset "absent", "directory", "file", "touch"
+$src = Get-AnsibleParam -obj $params -name "src" -type "path" -failifempty ($state -in @("hard", "link", "junction"))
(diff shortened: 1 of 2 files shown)


jborean93:
I would be very reluctant to using the pwsh filesystem provider for managing symbolic links. It's support for things like relative paths, or using raw NT device path targets is quite limited. It's especially problematic in future pwsh versions where it sometimes works and sometimes does not. Dealing with existing strings that are already the NT device path is also a mixed bag. Unfortunately the next best thing is `cmd.exe /c mklink ...` which has other problems, like getting the existing target for idempotency checks.

This means the only real way to have a reliable symbolic/junction handler in Ansible would be to use the PInvoke code itself. We actually have a shared util for this currently called [Ansible.ModuleUtils.Link](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/powershell/Ansible.ModuleUtils.LinkUtil.psm1) but it's probably due to be refactored into this collection and slightly rewritten. The sheer amount of changes and the general UI involved sort off tells me that this should be a completely separate module rather than in `win_file`.  I understand that the POSIX `file` module does both at the same time but I don't think being consistent there is worth the disadvantages. If you are happy to try and work on this I can definitely give you some pointers or help as you go along but it is a lot of work, especially the PInvoke code and setting up the tests.

PR #508, opened 2023-05-25
update private key check when exporting to pkcs12

gotit96:
##### SUMMARY
This proposed change, update private key check, when exporting to pkcs12, to handle the case where the ``PrivateKey`` property is null despite it being there.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_certificate_store

##### ADDITIONAL INFORMATION
[X509Certificate2.PrivateKey](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.privatekey) is not always set, depending the method used to generate key. it's also marked as obsolete by microsoft.

A way to check this
```paste below
> New-SelfSignedCertificate -DnsName test.local

   PSParentPath: Microsoft.PowerShell.Security\Certificate::LocalMachine\MY

Thumbprint                                Subject
----------                                -------
DC9D1037DCB465195F5F90400100B2759C3DF8DB  CN=test.local

> (Get-Item Cert:\LocalMachine\My\DC9D1037DCB465195F5F90400100B2759C3DF8DB).HasPrivateKey
True
> (Get-Item Cert:\LocalMachine\My\DC9D1037DCB465195F5F90400100B2759C3DF8DB).PrivateKey

> [Security.Cryptography.X509Certificates.RSACertificateExtensions]::GetRSAPrivateKey((Get-Item Cert:\LocalMachine\My\DC9D1037DCB465195F5F90400100B2759C3DF8DB)).Key

AlgorithmGroup     : RSA
Algorithm          : RSA
ExportPolicy       : AllowExport
Handle             : Microsoft.Win32.SafeHandles.SafeNCryptKeyHandle
IsEphemeral        : False
IsMachineKey       : True
KeyName            : te-a3d1cfb3-c570-4d46-b33a-75de1bd2a175
KeySize            : 2048
KeyUsage           : Decryption, Signing
ParentWindowHandle : 0
Provider           : Microsoft Software Key Storage Provider
ProviderHandle     : Microsoft.Win32.SafeHandles.SafeNCryptProviderHandle
UniqueName         : de19eccdfb0a8f913ccc1589b9b9c5ef_4adca362-26b7-4222-911c-fa932fda9d7a
UIPolicy           : System.Security.Cryptography.CngUIPolicy
```

### DIFF
--- a/plugins/modules/win_certificate_store.ps1
+++ b/plugins/modules/win_certificate_store.ps1
@@ -228,31 +228,6 @@ Function New-CertFile($module, $cert, $path, $type, $password) {
         "der" { [System.Security.Cryptography.X509Certificates.X509ContentType]::Cert }
         "pkcs12" { [System.Security.Cryptography.X509Certificates.X509ContentType]::Pkcs12 }
     }
-    if ($type -eq "pkcs12") {
-        $missing_key = $true
-        if ($cert.HasPrivateKey) {
-            switch ($cert.PublicKey.Oid.FriendlyName) {
-                "DSA" {
-                    $certPrivateKey = [Security.Cryptography.X509Certificates.DSACertificateExtensions]::GetDSAPrivateKey($cert[0]).Key
-                }
-                "RSA" {
-                    $certPrivateKey = [Security.Cryptography.X509Certificates.RSACertificateExtensions]::GetRSAPrivateKey($cert[0]).Key
-                }
-                "ECC" {
-                    $certPrivateKey = [Security.Cryptography.X509Certificates.ECDsaCertificateExtensions]::GetECDsaPrivateKey($cert[0]).Key
-                }
-                default {
-                    $module.FailJson("Cannot export cert with key as PKCS12 $($cert.PublicKey.Oid.FriendlyName) algorithm isn't supported")
-                }
-            }
-            if (($null -ne $certPrivateKey) -and ($certPrivateKey.ExportPolicy -ne [System.Security.Cryptography.CngExportPolicies]::None)) {
-                $missing_key = $false
-            }
-        }
-        if ($missing_key) {
-            $module.FailJson("Cannot export cert with key as PKCS12 when the key is not marked as exportable or not accessible by the current user")
-        }
-    }
 
     if (Test-Path -LiteralPath $path) {
         Remove-Item -LiteralPath $path -Force
@@ -275,7 +250,8 @@ Function New-CertFile($module, $cert, $path, $type, $password) {
         $cert_bytes = $file_encoding.GetBytes($cert_content)
     }
     elseif ($type -eq "pkcs12") {
-        $module.Result.key_exportable = [string]$certPrivateKey.ExportPolicy
+        # to ensure backwards compatibility, if $false Export will fail.
+        $module.Result.key_exportable = $true
     }
 
     if (-not $module.CheckMode) {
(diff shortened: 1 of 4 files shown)


plugins/modules/win_certificate_store.ps1
```diff
@@ -262,10 +275,7 @@ Function New-CertFile($module, $cert, $path, $type, $password) {
         $cert_bytes = $file_encoding.GetBytes($cert_content)
     }
     elseif ($type -eq "pkcs12") {
-        $module.Result.key_exported = $false
-        if ($null -ne $cert.PrivateKey) {
-            $module.Result.key_exportable = $cert.PrivateKey.CspKeyContainerInfo.Exportable
-        }
+        $module.Result.key_exportable = [string]$certPrivateKey.ExportPolicy
```
jborean93, line 278: This is changing the value around from being a bool to a string value. If you are wanting to add the export policy value as a return value we would need to add it as a new return value to ensure backwards compatibility.

plugins/modules/win_certificate_store.ps1
```diff
@@ -229,12 +229,25 @@ Function New-CertFile($module, $cert, $path, $type, $password) {
...
     if ($type -eq "pkcs12") {
-        $missing_key = $false
-        if ($null -eq $cert.PrivateKey) {
-            $missing_key = $true
-        }
-        elseif ($cert.PrivateKey.CspKeyContainerInfo.Exportable -eq $false) {
-            $missing_key = $true
+        $missing_key = $true
+        if ($cert.HasPrivateKey) {
+            switch ($cert.PublicKey.Oid.FriendlyName) {
```
jborean93, line 234: Is there any benefit to actually trying to export the private key bytes vs just relying on `HasPrivateKey`? I see you are checking the `ExportPolicy` later on but is that needed at all vs just relying on `Export()` to give us what we need and potentially failing?

Avoiding having to deal with the different key types would be nice plus these methods require the host to have dotnet framework 4.6 which isn't a current guarantee. If we do need these checks we should only optionally do them if the type/method is present in the runtime.

jborean93 (commented):
Just left a few comments. One question I do have, do you know if it's possible/easy to setup a certificate that uses CNG vs CAPI where this scenario would actually happen. It would be great for our CI runner to test this particular scenario going forward.

Thanks for the PR.

plugins/modules/win_certificate_store.ps1
```diff
@@ -262,10 +250,8 @@ Function New-CertFile($module, $cert, $path, $type, $password) {
         $cert_bytes = $file_encoding.GetBytes($cert_content)
     }
     elseif ($type -eq "pkcs12") {
-        $module.Result.key_exported = $false
-        if ($null -ne $cert.PrivateKey) {
-            $module.Result.key_exportable = $cert.PrivateKey.CspKeyContainerInfo.Exportable
-        }
+        # to ensure backwards compatibility, if $false Export will fail.
+        $module.Result.key_exportable = $true
```
jborean93, line 254: Why the change in this value? While it might still be `$false` for CNG keys it'll still be valid for CAPI ones. I prefer it to be sometimes right than always set to `$true`.

gotit96:
Hi,

Updated to rely on Export() to give us what need and potentially failing. And added few integration test uses CNG and CAPI key provider.

jborean93 (approved):

jborean93:
Thank you very much for the changes, especially adding the integration tests!

PR #519, opened 2023-06-19
Potential fix for issue #518

### LINKED ISSUE #518: Add post_reboot_delay option to win_updates module
##### SUMMARY
Adding a  "post_reboot_delay" option to the win_updates module. 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

win_updates

##### ADDITIONAL INFORMATION

This will help allowing sometime for the Windows system to settled after the automatic reboots during patching and thus avoiding unexpected behavior and issues (such as winrm send_input failures)  when the reboot option is used.

```yaml

```

johnjore:
##### SUMMARY

Post patching, and after reboot, WinRM connection is not stable. While new uptime is received, win_updates crashes with WinRMOperationTimeoutError when attempting another cycle of patching / updates and/or validating if there are more patches to install.

From log:
```
[WARNING]: ERROR DURING WINRM SEND INPUT - attempting to recover:
WinRMOperationTimeoutError
```

Potentially fixes #518 by allowing the connection to "sleep" for n seconds after connection is made, post reboot, before continuing. This allows the WinRM connection to be more reliable and avoids crashing win_updates. 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Adds support for post_reboot_delay parameter, from win_reboot module, to win_updates module

### DIFF
--- a/plugins/action/win_updates.py
+++ b/plugins/action/win_updates.py
@@ -710,12 +710,14 @@ class ActionModule(ActionBase):
         'reboot',
         'skip_optional',
         'reboot_timeout',
+        'post_reboot_delay',
         'reject_list',
         'server_selection',
         'state',
     ]
 
     DEFAULT_REBOOT_TIMEOUT = 1200
+    DEFAULT_POST_REBOOT_DELAY = 0
 
     def __init__(self, *args, **kwargs):
         super(ActionModule, self).__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


jborean93:
Sorry I haven't been ignoring this but trying to think off what's the best thing to do here. Setting an arbitrary wait is not ideal as it can't guarantee that the wait is going to be long enough and even if it is, the task has been delayed even further. The real fix is going to be a lot more complicated. Essentially we are going to be a retry mechanism and a way to catch this particular error.

johnjore:
I dont disagree, its anything but ideal, but I suspect there is a reason win_reboot has this parameter. 

The RH consultant re-wrote bits of the playbook and tasks so win_update no longer does the reboot, but calls win_reboot directly, with this delay set, as a 2nd task. It then has to implement the whole "if there are still patches to be installed", loop around again, while also ignoring errors from win_update etc. Something that was 10 lines became 30-40 as this delay parameter was missing.

Short term: Support for this parameter is needed to keep things clean, and readable. Long term solution is to resolve/mask/fix/workaround the root cause of "WinRMOperationTimeoutError's.

jborean93:
> but I suspect there is a reason win_reboot has this parameter.

Mostly it's just a historical thing, plus `win_reboot` has had to help people handle flaky reboots for tasks that occur afterwards. In the `win_updates` case it can do all that internally without having the user explicitly opt into it. I agree the workaround proposed by the RH consultant is not ideal and `reboot: true` should be improved to handle these types of cases but I unfortunately don't think this particular PR is the way to go sorry.

PR #534, opened 2023-07-29
Fix broken path in case of volume junction

gillg:
##### SUMMARY
win_acl is bronken in case of junction + follow option

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_acl

##### ADDITIONAL INFORMATION
By using the option `folllow`, if the link is a junction to a volume we get an error like :
`"Volume{ad5b0320-77a0-4523-a8c8-00a52b860819}\\ file or directory does not exist on the host"`

The internal function `Get-Link` seems returns the absolute UNC path without prefix...

### DIFF
--- a/plugins/modules/win_acl.ps1
+++ b/plugins/modules/win_acl.ps1
@@ -116,7 +116,7 @@ while ($follow) {
 
     if ($link_info -and $link_info.Type -in @("SymbolicLink", "JunctionPoint")) {
         $path = $link_info.AbsolutePath
-        if ($link_info.Type -eq "JunctionPoint") {
+        if ($link_info.SubstituteName -like "\??\*") {
             $path = "\\?\" + $path
         }
     }
--- a/plugins/modules/win_acl.ps1
+++ b/plugins/modules/win_acl.ps1
@@ -116,6 +116,9 @@ while ($follow) {
 
     if ($link_info -and $link_info.Type -in @("SymbolicLink", "JunctionPoint")) {
         $path = $link_info.AbsolutePath
+        if ($link_info.Type -eq "JunctionPoint") {
+            $path = "\\?\" + $path
+        }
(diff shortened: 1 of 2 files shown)


plugins/modules/win_acl.ps1
```diff
@@ -116,6 +116,9 @@ while ($follow) {
 
     if ($link_info -and $link_info.Type -in @("SymbolicLink", "JunctionPoint")) {
         $path = $link_info.AbsolutePath
+        if ($link_info.Type -eq "JunctionPoint") {
```
jborean93, line 119: I'm not sure this is the correct check to do here. While a junction point might refer directly to an NT device/volume it can still be used to point to a folder like a symbolic link. It might be worthwhile to check `$link_info.SubstituteName` which is the raw path of the junction point target and see if it's in the `\??\Volume{guid}` format. It could also check if the `$path` starts with `Volume{guid}` pattern.

Only then should the new code here prefix the `AbsolutePath` with `\\?\`.

jborean93 (commented):
Just left a comment around only prefixing `\\?\` if it's an actual volume path. We also would need to see a changelog fragment for this bugfix https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment.

plugins/modules/win_acl.ps1
gillg: Note for info:
`$link_info.SubstituteName = \??\Volume{f7ae69d2-76fe-4650-b4ba-9afab7c8dfeb}\`
`$path = Volume{f7ae69d2-76fe-4650-b4ba-9afab7c8dfeb}\`

I don't know yet what is the best approach to be more generic as possible

jborean93 (approved):

Issue #530, opened 2023-07-17
Windows update KB2267602 keeps reinstalling causing loop.   


GDMTT:
##### SUMMARY
There is currently an issue with KB2267602 where it installs and then shows up as available to install again and no matter how many times it's installed it is still listed as available for install. This has the effect of causing win update to loop, which blocks the progress  in the job. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win update

##### ANSIBLE VERSION
bash-5.1$ ansible --version
ansible [core 2.15.2rc1]
  config file = None
  configured module search path = ['/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  ansible collection location = /runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.9.17 (main, Jun 26 2023, 00:00:00) [GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
bash-5.1$ 

```

##### COLLECTION VERSION
# /usr/share/ansible/collections/ansible_collections
Collection              Version
----------------------- -------
amazon.aws              6.2.0  
ansible.posix           1.5.4  
ansible.windows         2.0.0  
awx.awx                 22.5.0 
azure.azcollection      1.16.0 
community.vmware        3.8.0  
google.cloud            1.2.0  
kubernetes.core         2.4.0  
openstack.cloud         2.1.0  
ovirt.ovirt             3.1.2  
redhatinsights.insights 1.0.7  
theforeman.foreman      3.12.0 

note ansible-galaxy collection list community.general returns nothing. 

##### CONFIGURATION
ash-5.1$ ansible-config dump --only-changed
CONFIG_FILE() = None
bash-5.1$ 

##### OS / ENVIRONMENT
AWX 22.3.0
AWX-EE (latest)

##### STEPS TO REPRODUCE
Playbook used:
---
 - hosts: all
   serial:
     - 5
   tasks:
   - name: Ping host to wake up
     ansible.windows.win_ping:

   - name: Check for missing updates
     win_updates:
       state: searched
     register: update_results

   - name: report update results
     debug:
       msg: |
         {% for k in update_results.updates %}
         {{ update_results.updates[k].title }}
         {% endfor %}         

[...]

Issue #525, opened 2023-07-05
win_update: Ansible and the operating system think that the updates are installed, in fact they are not.


valentusys:
##### SUMMARY
There are 2 simple playbooks - one checks for updates and displays a list of them, the second installs all available updates.

Both playbooks work correctly - no errors.
However, when searching for updates again, they are found again. This is confirmed by the display in the Windows GUI.

At the same time, the update service logs report that the updates are installed

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION
ansible [core 2.12.10]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 2.11.3
  libyaml = True

```

##### COLLECTION VERSION

# /usr/lib/python3/dist-packages/ansible_collections
Collection      Version
--------------- -------
ansible.windows 1.10.0

# /root/.ansible/collections/ansible_collections
Collection      Version
--------------- -------
ansible.windows 2.0.0

```

##### CONFIGURATION
```
CALLBACKS_ENABLED(/etc/ansible/ansible.cfg) = ['timer', 'mail']
DEFAULT_FORKS(/etc/ansible/ansible.cfg) = 50
DEFAULT_GATHER_TIMEOUT(/etc/ansible/ansible.cfg) = 30
DEFAULT_LOG_PATH(/etc/ansible/ansible.cfg) = /var/log/ansible.log
DEFAULT_STDOUT_CALLBACK(env: ANSIBLE_STDOUT_CALLBACK) = debug
HOST_KEY_CHECKING(/etc/ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/etc/ansible/ansible.cfg) = auto_silent

```

##### OS / ENVIRONMENT
Host: debian 11
Target: Windows 10 Pro(10.0.18363)
Target: Windows Server 2022 Standard(10.0.20348)

##### STEPS TO REPRODUCE

ansible-playbook /root/ansible/win-servers/win-update-search.yaml -i /root/ansible/win-servers/inventory/test

```
win-update-search.yaml
---
[...]

Issue #343, opened 2022-03-20
win_update reboot on failure


forum4cc:
##### SUMMARY
Sometimes Windows Updates are failing because of multiple reasons. For example with the error code 0x8024a204. In most of the cases, a reboot helps, sometimes you need to clear the update cache aswell. If you loop over the updates with "until:" you can't catch the error with a reboot, so this must be implemented into win_updates directly.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
win_update

##### ADDITIONAL INFORMATION
Add another option, for example "reboot_on_failure: yes/no". If win_update has a failed_update_count >= 1 and reboot_on_failure is set to true, reboot the machine.

Current state, when an update fails, you are left without updates:
```
---
- name: install windows updates and reboot until there are no more updates left
  ignore_errors: yes
  win_updates:
    category_names: '*'
    reboot: yes
    state: installed
  register: win_updates
  until: win_updates.installed_update_count == 0
```

With the feature request the update fails, it reboots the machine and hopefully installs the update within the next retry.
```
---
- name: install windows updates and reboot until there are no more updates left
  ignore_errors: yes
  win_updates:
    category_names: '*'
    reboot: yes
    state: installed
    reboot_on_failure: yes
  register: win_updates
  until: (win_updates.installed_update_count == 0 ) and (win_updates.failed_update_count == 0)
```
There must be a limit to this aswell to not get stuck in a loop if rebooting doesn't help to solve the problem.

A workaround for this might restart the machine before starting the updates, but somehow Windows sometimes needs a failing installation before rebooting helps.

jborean93:
Are you abel to share what Ansible version or `ansible.windows` version you are running with. The `win_updates` plugin has been rewritten in recent versions to try and handle failures such as this. If `reboot: yes` is set then it should be attempting a reboot on a failure and trying again as per https://github.com/ansible-collections/ansible.windows/blob/108281ac99dc89e7500acf8c6a1695d706f93bee/plugins/action/win_updates.py#L851-L857. It will reboot once on a failure and if that didn't fix the problem then it will return the failure back to the user to handle accordingly.

gillg:
I think I have the same issue here !
On the point of view of the action plugin there is no reason to break the `while True` loop and there is no option to limit the round number.
I'm wondering if there is no something wrong in this specific microsoft update (KB5011503), but anyway we should have a way to break the update loop after a certain number of rounds.

For the root cause, I don't have any idea.

[...]

PR #542, opened 2023-08-14
win_updates - fix reboot loop behaviour

jborean93:
##### SUMMARY
Fixes up the logic in the action plugin to not retry if the same update was found to be installed. This should pickup scenarios where an update was rolled back during a reboot or WUA failed to detect the install failed to install.

Also adds extra handling for a failure of KB2267602 which contains Microsoft Defender definitions to try and install a failed update using the Defender executable. This has shown to be more reliable than WUA when it comes to installing this update.

Fixes: https://github.com/ansible-collections/ansible.windows/issues/343
Fixes: https://github.com/ansible-collections/ansible.windows/issues/525
Fixes: https://github.com/ansible-collections/ansible.windows/issues/530

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_updates


plugins/action/win_updates.py
```diff
@@ -756,6 +757,32 @@ def _run_sync(self, task_vars, module_options, reboot, reboot_timeout):  # type:
...
+            # is to detect an update that may have been rolled back in the last
+            # reboot or whether WUA failed to report that the update failed.
+            same_updates = len(update_result.install_results) > 0
+            current_updates = set()
+            for update_id in update_result.install_results.keys():
+                if update_id in installed_updates:
+                    current_updates.add(update_id)
+                else:
+                    same_updates = False
+                    installed_updates.add(update_id)
```
nitzmahone, line 770: Since we've already got sets, isn't this better expressed in basic set operations, something like:

```suggestion
            current_updates = set(update_result.install_results.keys())
            duplicated_updates = current_updates.intersection(installed_updates)
            installed_updates.update(current_updates)
            
            if duplicated_updates:
            ...
```

Unless I'm misunderstanding the logic, the existing imperative code will also be "blind" to any duplicate that occurs that isn't the first non-duplicate, right? I don't think that's the desired behavior, so doing it all with sets also saves you from that bug.

plugins/modules/win_updates.ps1
```diff
@@ -1909,17 +1909,57 @@ namespace Ansible.Windows.WinUpdates
     $installResult = Invoke-AsyncMethod 'Installing updates' $api.InstallAsync($installer, ${function:Receive-CallbackProgress}, $CancelToken)
     $exitResult.changed = $true
 
+    # https://www.microsoft.com/en-us/wdsi/defenderupdates
+    $defenderExe = [System.IO.Path]::Combine($env:ProgramFiles, 'Windows Defender', 'MpCmdRun.exe')
```
nitzmahone, line 1913: Would it make more sense to move all this stuff down inside the conditional block where the work is actually done (including deferring the Defender presence detection), so it's all self-contained and only being done when the update in question was included and the other conditions are met? Feels wrong to have it always running.

plugins/modules/win_updates.ps1
```diff
@@ -1909,17 +1909,57 @@ namespace Ansible.Windows.WinUpdates
...
+        # false positive causing an infinite loop. The MpCmdRun command can be
+        # used to install the update outside of WUA which seems to work when
+        # WUA does not.
+        if (
+            $resultCode -ne 'Succeeded' -and
+            '2267602' -in $updateKBs -and
+            (Test-Path -LiteralPath $defenderExe)
+        ) {
+            $null = & $runDefenderCommand '-RemoveDefinitions' '-DynamicSignatures'
+            $rc = & $runDefenderCommand '-SignatureUpdate'
```
nitzmahone, line 1954: If the `-RemoveDefinitions` succeeds and the `-SignatureUpdate` fails, won't we have temporarily broken Defender with no warning/failure from the module itself?

nitzmahone (changes requested):
Good times on this one- not... A couple of maintainability suggestions and a question about intermediate failure on the Defender recovery, but overall LGTM.

plugins/modules/win_updates.ps1
jborean93: I don't think so, the help for `MpCmdRun.exe` has

```
   -RemoveDefinitions
        Restores the last set of signature definitions

        [-All]
        Removes any installed signature and engine files. Use this
        option if you have difficulties trying to update signatures.

        [-DynamicSignatures]
        Removes all Dynamic Signatures.
```

I can't verify for sure but it sounds like this just ensures the latest working set will be restored giving it a stable base and potentially removing any cached updates. Unfortunately these are things that are hard to verify as I cannot really induce this scenario.

plugins/modules/win_updates.ps1
jborean93: It just means I need to redefine the exe in every loop, this was it's done once and the helper code below is just a scriptblock that is only run in the scenario in question.

plugins/action/win_updates.py
jborean93: I've got the current logic so it's only if there are just duplicate updates. If it installed an update that hadn't been installed before it will still consider this round successful and try one more. That way if the current update somehow caused another to fail it will still try it by itself before considering it a lost cause.

plugins/modules/win_updates.ps1
nitzmahone: ah ok, cool- just wanted to make sure it wasn't like "zap the whole thing" :laughing:

plugins/action/win_updates.py
nitzmahone: ah ok, so maybe like:

```suggestion
            current_updates = set(update_result.install_results.keys())
            new_updates = current_updates.difference(installed_updates)
            installed_updates.update(current_updates)
            
            if current_updates and not new_updates:
            ...
```

I'm probably being overly pedantic- it just feels more readable to me since the logic is purely set-oriented anyway.

jborean93: Yea makes sense, I'll swap it over to this new logic.

plugins/modules/win_updates.ps1
jborean93: I means having to do an if check to see if it failed then another if check to see if the exe exists like

```powershell
if ($failedLogic) {
    $defenderExe = [System.IO.Path]::Combine(...)
    if (Test-Path $defenderExe) {
        do defender stuff
    }
}
```

I don't hate it but it's even more indentation inside an already indented section.

nitzmahone: No worries- might want to revisit abstracting a more generic "recovery tweaks for specific known problems" mechanism if there end up being more of these kinds of things (:crossed_fingers: for not), but totally fine as-is.

nitzmahone (approved):

PR #574, opened 2023-12-13
New win_assemble module

DanielTOsborne:
##### SUMMARY

I've created a module to combine multiple text files into a single one, like assemble.

We have an application that uses INI files for managing its configuration with many field deployments. I wanted to centrally manage their configuration, with each INI section contained in its own file centrally, but combined into a single file when deployed.

The ansible.builtin.assemble module accomplishes this goal, but only for unix targets. Our use case uses Windows targets, so a new module needed to be written to accomplish this.

The goal was to implement a module with feature parity to ansible.built.in.assemble, mimicking its behavior and config options, as closely as possible.

One area where I diverged from the builtin.assemble, was I added the option for a header and footer in the generated output. This is useful to indicate to readers that the file is generated and edits can be lost. See my example below.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_assemble

##### ADDITIONAL INFORMATION

```yaml
  - name: test
    ansible.windows.win_assemble:
      remote_src: false
      src: "conf/DCPs/{{inventory_hostname}}"
      dest: "{{dcs_root}}\\config.test.ini"
      delimiter: '#'
      backup: true
      decrypt: true
      regexp: "ini$"
      header: '#### THIS FILE IS GENERATED. CHANGES MAY BE LOST! ####'
```

### DIFF
--- a/plugins/action/win_assemble.py
+++ b/plugins/action/win_assemble.py
@@ -1,0 +1,161 @@
+# (c) 2013-2016, Michael DeHaan <michael.dehaan@gmail.com>
+#           Stephen Fromm <sfromm@gmail.com>
+#           Brian Coca  <briancoca+dev@gmail.com>
+#           Toshio Kuratomi  <tkuratomi@ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


DanielTOsborne:
Note, this was written by an employee of the US Federal Government, for official government use. As such, any contributed content not already covered under copyright is thereby released under the Public Domain and is not copyrighted.

jborean93:
Thanks for the module submissions but unfortunately this is not something we are interested in having for this collection at this point in time. You are more than welcome to place this in your own collection or continue to use this internally but we have a high bar for acceptance with `ansible.windows` and `community.windows` is not currently accepting new plugins.

Sorry if this wasn't the news you wanted to hear, all the best for the future.

DanielTOsborne:
Oh, I saw community.windows wasn't accepting any, but I didn't see any mention that ansible.windows wasn't.

Certainly not the response I was expecting for sure. I don't really care personally. I've been using it internally for some time and just wanted to contribute back upstream.

Issue #579, opened 2024-02-01
ansible.windows.win_copy hits PermissionDenied UnauthorizedAccessException when copying a local file to a remote node when using become to runas an unprivileged user


ChrisNicoll-Pro:
##### SUMMARY
It is not possible to use `ansible.windows.win_copy` when running with `become` set to an unprivileged user that differs to the one used to connect to the remote node. The module fails with a `PermissionDenied` `UnauthorizedAccessException` error when copying the file from the remote temp directory to its final destination. Based on the error output, this looks to be because the temporary directory of the connection user is incorrectly chosen when the file is initially transferred to the remote node. As a result, the file ends up under a different user's home directory which an unprivileged user is unable to read from. The module falls over when trying to verify the existence of the file on the remote node https://github.com/ansible-collections/ansible.windows/blob/main/plugins/modules/win_copy.ps1#L431.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_copy

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.3]
  config file = /Users/****/work/ansible/ansible.cfg
  configured module search path = ['/Users/****/work/ansible/library']
  ansible python module location = /Users/****/Library/Python/3.10/lib/python/site-packages/ansible
  ansible collection location = /Users/****/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/****/Library/Python/3.10/bin/ansible
  python version = 3.10.11 (v3.10.11:7d4cc5aa85, Apr  4 2023, 19:05:19) [Clang 13.0.0 (clang-1300.0.29.30)] (/usr/local/bin/python3)
  jinja version = 3.1.3
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection      Version
--------------- -------
ansible.windows 2.2.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

**Control Node**
macOS Sonoma

**Target Node**
Windows 10

##### STEPS TO REPRODUCE

Spin up a Windows node with an administrator account (e.g. `admin`) and an unprivileged account (e.g. `user1`)
Setup a playbook to copy a file from the control node to the remote node making sure to configure Ansible to connect via the `admin` account. The `win_copy` module should be invoked using `become` to run as the `user1` account.

```yaml
---
- hosts: node1
  vars:
    ansible_connection: 'winrm'
    ansible_user: 'admin'
    ansible_password: 'secure_password'
  tasks:
[...]

PR #583, opened 2024-02-07
Expand notes in win_copy and become

jborean93:
##### SUMMARY
Adds an extra note to further clarify what needs to be done when using win_copy with a become user that is not a member of the Administrators group on the target Windows host.

Fixes: https://github.com/ansible-collections/ansible.windows/issues/579

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
win_copy

### DIFF
--- a/plugins/modules/win_copy.py
+++ b/plugins/modules/win_copy.py
@@ -91,9 +91,9 @@ notes:
   mechanism. If sending large files consider hosting them on a web service and
   using M(ansible.windows.win_get_url) instead.
 - When using become with a user that is not an Administrator, the remote temp
-  directory must be set to a one that is accessible to both the become and
+  directory must be set to one that is accessible to both the become and
   connection user with C(ansible_remote_tmp). See the examples for more
-  more information.
+  information.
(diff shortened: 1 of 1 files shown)


plugins/modules/win_copy.py
```diff
@@ -90,6 +90,10 @@
 - Because win_copy runs over WinRM, it is not a very efficient transfer
   mechanism. If sending large files consider hosting them on a web service and
   using M(ansible.windows.win_get_url) instead.
+- When using become with a user that is not an Administrator, the remote temp
+  directory must be set to a one that is accessible to both the become and
+  connection user with C(ansible_remote_tmp). See the examples for more
+  more information.
```
ChrisNicoll-Pro, line 96: That extra note is great! Just a couple grammatical bits worth tidying up.
```suggestion
- When using become with a user that is not an Administrator, the remote temp
  directory must be set to one that is accessible to both the become and
  connection user with C(ansible_remote_tmp). See the examples for more
  information.
```

PR #587, opened 2024-02-26
Added depth parameter to win_uri ConvertTo-Json call

thomasriera-akkodis:
Added depth parameter to win_uri ConvertTo-Json call

##### SUMMARY
When using the win_uri module to send a json object, win_uri uses the ConvertTo-Json method. 

This method has a currently  unused parameter called Depth which represents the maximum depth of nested children to parse.
(https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/convertto-json?view=powershell-7.4#-depth)

This parameter's default value is 2, so the method does not correctly convert json objects with a higher depth.

I set it to a default of 20 which should be enough for most cases, or at least more useful than the default of 2.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_uri.ps1

##### ADDITIONAL INFORMATION
To reproduce the bug : send a json object with more than 2 levels of nesting using win_uri. The method will seem to work but the sent object will be incomplete.

With this change, the module can correctly send json objects with up to 20 levels of nesting.

### DIFF
--- a/plugins/modules/win_uri.ps1
+++ b/plugins/modules/win_uri.ps1
@@ -183,7 +183,7 @@ $response_script = {
 $body_st = $null
 if ($null -ne $body) {
     if ($body -is [System.Collections.IDictionary] -or $body -is [System.Collections.IList]) {
-        $body_string = ConvertTo-Json -InputObject $body -Compress
+        $body_string = ConvertTo-Json -InputObject $body -Compress -Depth 20
     }
     elseif ($body -isnot [String]) {
         $body_string = $body.ToString()
--- a/changelogs/fragments/PR-587-Fix-win_uri-json-depth.yaml
+++ b/changelogs/fragments/PR-587-Fix-win_uri-json-depth.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - win_uri - Max depth for json object conversion used to be 2. Can now send json objects with up to 20 levels of nesting
--- a/tests/integration/targets/win_uri/tasks/main.yml
+++ b/tests/integration/targets/win_uri/tasks/main.yml
@@ -325,7 +325,7 @@
     - not status_code_check.changed
     - status_code_check.status_code == 202
 
-  - name: send JSON body with dict type
+- name: send JSON body with dict type
   win_uri:
     url: https://{{httpbin_host}}/post
     method: POST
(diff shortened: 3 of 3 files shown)


plugins/modules/win_uri.ps1
```diff
@@ -183,7 +183,7 @@ $response_script = {
 $body_st = $null
 if ($null -ne $body) {
     if ($body -is [System.Collections.IDictionary] -or $body -is [System.Collections.IList]) {
-        $body_string = ConvertTo-Json -InputObject $body -Compress
+        $body_string = ConvertTo-Json -InputObject $body -Compress -Depth 20
```
jborean93, line 186:
```suggestion
        # The body comes from Ansible so it is safe to exceed the normal depth of 2
        $body_string = ConvertTo-Json -InputObject $body -Compress -Depth 20
```

jborean93 (changes requested):
Thanks for the PR here, I've added a tiny suggestion that adds a comment explaining why it is ok to add the custom depth.

For us to merge this PR we do need 2 extra things here

+ A changelog fragment
+ Some tests for this scenario

The changelog fragment is documented under https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment, this feature would come under `minor_changes` and would be placed in https://github.com/ansible-collections/ansible.windows/tree/main/changelogs/fragments

The tests for `win_uri` are located at https://github.com/ansible-collections/ansible.windows/blob/main/tests/integration/targets/win_uri/tasks/main.yml. You would probably want something similar to https://github.com/ansible-collections/ansible.windows/blob/c207ed49ddf73f152f56817787824a2c4ba93226/tests/integration/targets/win_uri/tasks/main.yml#L328-L342 just with a more nested object structure as a new test.

Please let me know if you need any help with the above and I'm happy to point you in the right direction.

jborean93:
@thomasriera-akkodis do you mind if I try and push some changes to get the tests working, I'm hoping to include this in the next release of this collection which I'm planning on doing soon.

thomasriera-akkodis:
@jborean93

I think everything should be okay now. Happy to see this change coming soon, I'll be able to remove my dirty workaround from my servers.

jborean93 (approved):
Thanks for working on this, the changes look great!

PR #602, opened 2024-04-10
win_powershell - add secure parameters

jborean93:
##### SUMMARY
Added parameters_secure_string to the win_powershell module that allows users to provide a SecureString or PSCredential object that is set with no_log.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_powershell

### DIFF
--- a/changelogs/fragments/win_powershell-securestring.yml
+++ b/changelogs/fragments/win_powershell-securestring.yml
@@ -1,0 +1,4 @@
+minor_changes:
+  - >-
+    win_powershell - Added the ``parameters_secure_string`` option that can be used to pass in a SecureString or
+    PSCredential parameter value.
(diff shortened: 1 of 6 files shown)


nitzmahone:
I think I'm starting to get the reason for this, but it does feel kinda brittle at a glance, and since the cleartext still enters the module, I'm wondering if there's a different thing we should be looking at to more securely channel those bits into the module and/or over the wire...

jborean93:
Without controller side argspec there isn't much we can do to achieve such a thing and it would also mean we need to implement some crypto mechanism on the Ansible side and the remote to achieve such a thing. The `psrp` connection plugin can provide a `SecureString` or `PSCredential` natively but any other connection plugin cannot which means rolling our own mechanism which means more round trips to negotiate the encryption used and probably sets us up for a CVE if we do it wrong.

The main reason for this PR is it adds the sensitive values to a parameter marked as `no_log` so users don't need to `no_log: True` the whole task invocation which makes things harder to debug as everything would be masked at runtime. Currently there is no real way of achieving this and even then things like `win_shell/win_command` are dangerous because of process auditing could log the sensitive values and with the existing `parameters` option scriptblock logging/transcription could expose the values if not passed in as a SecureString/PSCredential.

PR #643, opened 2024-08-13
Added Zone.Identifier mark for file downloaded with win_get_url module

JonnyShuali:
##### SUMMARY
Added the ability to set `Zone.Identifier` file stream for files downloaded using  _win_get_url_ .
This might be necessary for security purposes, file having this stream can be scanned using Windows Smartscreen and other malware analysis. Also, different policies can be set for files downloaded from the internet.  

Added the `set_zone` **optional** boolean parameter to the _win_get_url_ . Nothing changed on the output.

##### Example input
Before:
```
- name: Download src with sha256 checksum url
  ansible.windows.win_get_url:
    url: http://www.example.com/earthrise.jpg
    dest: C:\temp\earthrise.jpg
    checksum_url: http://www.example.com/sha256sum.txt
    checksum_algorithm: sha256
    force: true
```

After:
```
- name: Download src with sha256 checksum url
  ansible.windows.win_get_url:
    url: http://www.example.com/earthrise.jpg
    dest: C:\temp\earthrise.jpg
    checksum_url: http://www.example.com/sha256sum.txt
    checksum_algorithm: sha256
    force: true
    set_zone: true
```

### DIFF
--- a/plugins/modules/win_get_url.ps1
+++ b/plugins/modules/win_get_url.ps1
@@ -19,6 +19,7 @@ $spec = @{
         checksum = @{ type = 'str' }
         checksum_algorithm = @{ type = 'str'; default = 'sha1'; choices = @("md5", "sha1", "sha256", "sha384", "sha512") }
         checksum_url = @{ type = 'str' }
+        set_zone = @{ type = 'bool'; default = $false }
 
         # Defined for ease of use and backwards compatibility
         url_method = @{
@@ -41,6 +42,7 @@ $force = $module.Params.force
 $checksum = $module.Params.checksum
 $checksum_algorithm = $module.Params.checksum_algorithm
 $checksum_url = $module.Params.checksum_url
+$set_zone = $module.Params.set_zone
 
 $module.Result.elapsed = 0
 $module.Result.url = $url
@@ -144,14 +146,29 @@ Function Get-Checksum {
     }
     return $hash
 }
+Function Set-FileZoneStream {
+    <#
+    .SYNOPSIS
+    Set the input file with Zone.Identifier stream as downloaded from a browser.
+    #>
+    param(
+        [Parameter(Mandatory = $true)][Uri]$Url,
+        [Parameter(Mandatory = $true)][String]$Dest
+    )
+    if (-not (Test-Path -LiteralPath $Dest)) {
+        $module.FailJson("Cannt find '$Dest' for setting stream.")
+    }
+    Set-Content -Path $Dest -Stream Zone.Identifier -Value '[ZoneTransfer]', 'ZoneId=3', "ReferrerUrl=$Url", "HostUrl=$Url"
 
+}
 Function Invoke-DownloadFile {
     param(
         [Parameter(Mandatory = $true)][Ansible.Basic.AnsibleModule]$Module,
         [Parameter(Mandatory = $true)][Uri]$Uri,
         [Parameter(Mandatory = $true)][String]$Dest,
         [String]$Checksum,
-        [String]$ChecksumAlgorithm
+        [String]$ChecksumAlgorithm,
+        [bool]$SetZone = $false
     )
 
     # Check $dest parent folder exists before attempting download, which avoids unhelpful generic error message.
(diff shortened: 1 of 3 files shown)


plugins/modules/win_get_url.ps1
```diff
@@ -144,14 +146,30 @@ Function Get-Checksum {
...
+    #>
+    param(
+        [Parameter(Mandatory = $true)][Uri]$Url,
+        [Parameter(Mandatory = $true)][String]$Dest
+    )
+    if (-not (Test-Path -LiteralPath $Dest)) {
+        $module.FailJson("Cannt find '$Dest' for setting stream.")
+    }
+    $stream_dest = $Dest + ":Zone.Identifier"
+    Set-Content -Path $stream_dest -Value '[ZoneTransfer]', 'ZoneId=3', "ReferrerUrl=$Url", "HostUrl=$Url"
```
jborean93, line 162: Does a stream work with a `-LiteralPath` parameter? If so we will want to use that here.

plugins/modules/win_get_url.ps1
```diff
@@ -144,14 +146,30 @@ Function Get-Checksum {
...
+}
 Function Invoke-DownloadFile {
     param(
         [Parameter(Mandatory = $true)][Ansible.Basic.AnsibleModule]$Module,
         [Parameter(Mandatory = $true)][Uri]$Uri,
         [Parameter(Mandatory = $true)][String]$Dest,
         [String]$Checksum,
-        [String]$ChecksumAlgorithm
+        [String]$ChecksumAlgorithm,
+        [bool]$SetZone = $false
```
jborean93, line 172: We should use a switch here.

plugins/modules/win_get_url.ps1
```diff
@@ -201,6 +219,9 @@ Function Invoke-DownloadFile {
 
         if ($download) {
             Copy-Item -LiteralPath $tmp_dest -Destination $Dest -Force -WhatIf:$Module.CheckMode > $null
+            if ($SetZone) {
```
jborean93, line 222: We probably want to skip if in `$Module.CheckMode` as the file may not actually exist and we don't want to make a change to it anyway in check mode.

plugins/modules/win_get_url.ps1
```diff
@@ -264,14 +285,14 @@ if ($force -or -not (Test-Path -LiteralPath $dest)) {
     # force=yes or dest does not exist, download the file
     # Note: Invoke-DownloadFile will compare the checksums internally if dest exists
     Invoke-DownloadFile -Module $module -Uri $url -Dest $dest -Checksum $checksum `
-        -ChecksumAlgorithm $checksum_algorithm
+        -ChecksumAlgorithm $checksum_algorithm -SetZone $set_zone
```
jborean93, line 288: When using a switch we can do `-SetZone:$set_zone` to set it based on the module option.

plugins/modules/win_get_url.py
```diff
@@ -67,6 +67,12 @@
   url_timeout:
     aliases:
     - timeout
+  set_zone:
+    description:
+      - Markes the file as a file downloaded from internet zone for security purpose
+        using Zone.Identifier stream.
+    type: bool
+    default: no
```
jborean93, line 75: Will need a `version_added: 2.5.0` to document when the option was added. I also believe our linting wants boolean values to be `false` and not `no`.

plugins/modules/win_get_url.py
```diff
@@ -125,6 +131,13 @@
     checksum: a97e6837f60cec6da4491bab387296bbcd72bdba
     checksum_algorithm: sha1
     force: true
+
+- name: Download src with set_zone on
```
jborean93, line 135: Are you able to change with set_zone on to some a bit more descriptive, like download src and add the mark of the web or something along those lines?

jborean93 (commented):
Thanks for the great PR, I've added in a few comments. Let me know if you have any questions or anything else you need help with.

Are you also able to add a changelog fragment for this change as documented under https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments. The section should be a `minor_change` here.

PR #658, opened 2024-09-13
Update win_command and win_shell add `output_encoding_override` example

j796160836:
##### SUMMARY

Update `win_command.py` and `win_shell.py` add `output_encoding_override` example
I'm fixing ansible output encoding issue recently. During google search and ai answers which is not accurate.
Adding some example and trying to complete manual.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

- `ansible.windows.win_command` module
- `ansible.windows.win_shell` module

##### ADDITIONAL INFORMATION

Some related issue
https://github.com/ansible/ansible/issues/54896
https://github.com/ansible/ansible/pull/54966

### DIFF
--- a/plugins/modules/win_command.py
+++ b/plugins/modules/win_command.py
@@ -117,8 +117,8 @@ EXAMPLES = r'''
       - argument 1
       - -force
 
-- name: Run a batch script with big5 encoding (fixing garbled text issue)
-  ansible.windows.win_command: C:\somescript.bat
+- name: Run an executable that outputs text with big5 encoding
+  ansible.windows.win_command: C:\someprog.exe
   args:
     output_encoding_override: big5
 '''
--- a/plugins/modules/win_command.py
+++ b/plugins/modules/win_command.py
@@ -116,6 +116,11 @@ EXAMPLES = r'''
       - C:\Program Files\My Application\run.exe
       - argument 1
       - -force
+
+- name: Run a batch script with big5 encoding (fixing garbled text issue)
+  ansible.windows.win_command: C:\somescript.bat
+  args:
+    output_encoding_override: big5
 '''
 
 RETURN = r'''
--- a/plugins/modules/win_shell.py
+++ b/plugins/modules/win_shell.py
@@ -115,6 +115,11 @@ EXAMPLES = r'''
   ansible.windows.win_shell: C:\somescript.ps1
   args:
     no_profile: true
+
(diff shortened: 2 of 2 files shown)


plugins/modules/win_command.py
```diff
@@ -116,6 +116,11 @@
       - C:\Program Files\My Application\run.exe
       - argument 1
       - -force
+
+- name: Run a batch script with big5 encoding (fixing garbled text issue)
+  ansible.windows.win_command: C:\somescript.bat
+  args:
+    output_encoding_override: big5
```
jborean93, line 123:
```suggestion
- name: Run an executable that outputs text with big5 encoding
  ansible.windows.win_command: C:\someprog.exe
  args:
    output_encoding_override: big5
```

I would avoid mentioning it's a `.bat` script and explicitly set in the name that it outputs that encoding. Saying `(fixing garbled text issue)` may indicate to people that using `big5` will fix all those issues when in reality it's there for the rare cases when an executable doesn't use the existing console encoding (which we set to utf-8).

plugins/modules/win_shell.py
```diff
@@ -115,6 +115,11 @@
   ansible.windows.win_shell: C:\somescript.ps1
   args:
     no_profile: true
+
+- name: Run a PowerShell script with big5 encoding (fixing garbled text issue)
```
jborean93, line 119: I'm not sure there is really a need to document this for `win_shell`. PowerShell will be using the existing console encoding which we set to `utf-8`. Any executable that the `.ps1` script will call is subject to the encoding rules of PowerShell itself and shouldn't affect the `output_encoding_override` unless the script itself changes the `[Console]::OutputEncoding`.

j796160836: I'm a user for Taiwan, using Traditional Chinese.
Tested using Windows 2012 R2 as managed nodes.

plugins/modules/win_command.py
j796160836: I think it's common case in non-English systems, especially in Chinese, Japanese.

jborean93 (approved):

plugins/modules/win_shell.py
jborean93: Thanks for sharing, I think this is the case because keytool will be writing directly to the console and not going through PowerShell's string encoding. Appreciate you sharing this use case as I wasn't aware that would have been a problem.

PR #687, opened 2024-11-03
Migrate win_certificate_info module to ansible.windows repo from ansible.community repo

ronger4:
##### SUMMARY

 - Migration of win_certificate_info module from community repo to ansible.windows repo.
 - Added extra integration test for store_location: CurrentUser.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_certificate_info
##### ADDITIONAL INFORMATION

After this PR is pushed, we need to add a deprecation message in the documentation of win_certificate_info module in ansible.community

```paste below

```


tests/integration/targets/win_certificate_data/tasks/tests.yml
```diff
@@ -0,0 +1,102 @@
...
+      - root_stats.exists
+      - root_stats.certificates[0].archived == false
+      - root_stats.certificates[0].dns_names == ["root.ansible.com"]
+      - root_stats.certificates[0].extensions | count == 3
+      - root_stats.certificates[0].has_private_key == false
+      - root_stats.certificates[0].issued_by == "root.ansible.com"
+      - root_stats.certificates[0].issued_to == "root.ansible.com"
+      - root_stats.certificates[0].issuer == "C=US, CN=root.ansible.com"
+      - root_stats.certificates[0].path_length_constraint == 0
+#      - root_stats.certificates[0].public_key == (root_pub.stdout_lines | join())
```
Yaish25491, line 51: Can this be cleaned up? If not, please add a comment on why it's needed. 
This applied to row 80 as well

Yaish25491 (commented):
Looks grate, Added 1 minor comment.

tests/integration/targets/win_certificate_data/tasks/tests.yml
ronger4: It was commented originally, so can't tell why. But I will remove it as it's an old code anyways and if it wasn't added back, it seems like it can be removed.

plugins/modules/win_certificate_info.py
```diff
@@ -0,0 +1,228 @@
...
+
+# Copyright: (c) 2016, Ansible, inc
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_certificate_info
+short_description: Get information on certificates from a Windows Certificate Store
+description:
+- Returns information about certificates in a Windows Certificate Store.
```
jborean93, line 12: We need to set the version when this module was added to the collection through the `version_added` key.

```suggestion
- Returns information about certificates in a Windows Certificate Store.
version_added: 2.6.0
```

plugins/modules/win_certificate_info.py
```diff
@@ -0,0 +1,228 @@
...
+    type: str
+    choices: [ CurrentUser, LocalMachine ]
+    default: LocalMachine
+seealso:
+- module: ansible.windows.win_certificate_store
+author:
+- Micah Hunsberger (@mhunsber)
+'''
+
+EXAMPLES = r'''
```
jborean93, line 40: All these examples need to update `community.windows.win_certificate_data` to `ansible.windows.win_certificate_info`.

plugins/modules/win_certificate_info.py
ronger4: The reason you put 2.6.0 because it will be the upcoming version release, right?

Is there a place where we can follow the planned release dates if there is such

jborean93: The version is in the [galaxy.yml](https://github.com/ansible-collections/ansible.windows/blob/main/galaxy.yml) file and while sometimes it may have already been bumped it may still track the older release. I just compare it with the [tagged releases](https://github.com/ansible-collections/ansible.windows/releases) and see if I need to increment it or not. In this case `2.5.0` is the latest release and the `galaxy.yml` has not been updated yet. A minor version bump is needed for new features so the next release will be `2.6.0`.

As for a release schedule there hasn't been a hard rule of when to release. I usually aim for a release every 1-2 months but that is also dependent on whether there are changes to be made in a release.

jborean93 (approved):

jborean93:
Thanks, let's hold off for now on deprecating the `community.windows` variant as we need to sort out the next release timings and what other ones we want to migrate across for the next release. We can do the deprecations at the end.

PR #697, opened 2024-11-19
win_copy documentation fix - added flag `

marwin1991:
##### SUMMARY

Added information about flag `changed` for win_copy

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/win_copy.py
+++ b/plugins/modules/win_copy.py
@@ -162,6 +162,10 @@ EXAMPLES = r"""
 """
 
 RETURN = r"""
+changed:
+    description: Status of copy file execution inforing about destination file changes
+    type: bool
+    sample: true
 backup_file:
     description: Name of the backup file that was created.
     returned: if backup=yes


jborean93:
The `changed` value is a given for a module and it being true means that the module did an action, in this case a copy occurred. We don't typically report on the common return values like `changed` because all modules have them and it means the same thing. Was there something else you thought `changed: true` meant in the return value?

marwin1991:
@jborean93 I am new to ansible, so during creating some basic script with 'when:' condition I was researching dosc, using AI, but cloud not find anything. I was using some crazy contributions with backup_files variable, but then I decide to debug the output and I saw 'changed' property, and I thought maybe some else is struggling with it. Also AI having better docs could provide solution? I don't know, it's your decision, IMO giving as much info as we can simplifies usage of things :)

jborean93:
`changed` is just one of those things always there, if a module reported a change then it did what it was meant to do. I don't like to discount the work you've done to open the PR to try and improve the docs but in this case I might have to sorry. I do very much appreciate you contributing back and trying to improve the docs though.

Also just an FYI here is a doc on common return values from modules https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html.

jborean93:
> sorry for wasting your time

Definitely didn't waste my time, always good to see contributions. I'm just sorry it wasn't something we are looking to merge.

PR #705, opened 2024-11-27
win_timezone - Upgrade to Ansible.Basic style

jborean93:
##### SUMMARY
Upgrades the win_timzone module to the Ansible.Basic style module. This cleans up the older legacy module_util code and aligns it with the newer standards we try to write new modules in.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_timezone

##### ADDITIONAL INFORMATION
Typically I would have to create a changelog fragment for this change but as this is a new module that has not been released yet there are no public facing changes for users so it does not need one.

### DIFF
--- a/plugins/modules/win_timezone.ps1
+++ b/plugins/modules/win_timezone.ps1
@@ -3,71 +3,75 @@
 # Copyright: (c) 2015, Phil Schwartz <schwartzmx@gmail.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-#Requires -Module Ansible.ModuleUtils.Legacy
+#AnsibleRequires -CSharpUtil Ansible.Basic
 
-$params = Parse-Args $args -supports_check_mode $true
-$check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "bool" -default $false
-$diff_support = Get-AnsibleParam -obj $params -name "_ansible_diff" -type "bool" -default $false
-
-$timezone = Get-AnsibleParam -obj $params -name "timezone" -type "str" -failifempty $true
-
-$result = @{
-    changed = $false
-    previous_timezone = $timezone
-    timezone = $timezone
-}
-
-Try {
-    # Get the current timezone set
-    $result.previous_timezone = $(tzutil.exe /g)
-    If ($LASTEXITCODE -ne 0) {
-        Throw "An error occurred when getting the current machine's timezone setting."
-    }
-
-    if ( $result.previous_timezone -eq $timezone ) {
-        Exit-Json $result "Timezone '$timezone' is already set on this machine"
-    }
-    Else {
-        # Check that timezone is listed as an available timezone to the machine
-        $tzList = $(tzutil.exe /l).ToLower()
-        If ($LASTEXITCODE -ne 0) {
-            Throw "An error occurred when listing the available timezones."
-        }
-
-        $tzExists = $tzList.Contains(($timezone -Replace '_dstoff').ToLower())
-        if (-not $tzExists) {
-            Fail-Json $result "The specified timezone: $timezone isn't supported on the machine."
-        }
-
-        if ($check_mode) {
-            $result.changed = $true
-        }
-        else {
-            tzutil.exe /s "$timezone"
-            if ($LASTEXITCODE -ne 0) {
-                Throw "An error occurred when setting the specified timezone with tzutil."
-            }
-
-            $new_timezone = $(tzutil.exe /g)
-            if ($LASTEXITCODE -ne 0) {
-                Throw "An error occurred when getting the current machine's timezone setting."
-            }
-
-            if ($timezone -eq $new_timezone) {
-                $result.changed = $true
-            }
-        }
-
-        if ($diff_support) {
-            $result.diff = @{
-                before = "$($result.previous_timezone)`n"
-                after = "$timezone`n"
-            }
+$spec = @{
+    options = @{
+        timezone = @{
+            required = $true
+            type = "str"
         }
     }
-}
-Catch {
-    Fail-Json $result "Error setting timezone to: $timezone."
+    supports_check_mode = $true
 }
 
-Exit-Json $result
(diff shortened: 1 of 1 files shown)


plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
 # Copyright: (c) 2015, Phil Schwartz <schwartzmx@gmail.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-#Requires -Module Ansible.ModuleUtils.Legacy
+#AnsibleRequires -CSharpUtil Ansible.Basic
```
jborean93, line 6: This newer requires statement is used by Ansible to load in the `Ansible.Basic` C# module util.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
 
-$timezone = Get-AnsibleParam -obj $params -name "timezone" -type "str" -failifempty $true
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
 
-$result = @{
-    changed = $false
-    previous_timezone = $timezone
-    timezone = $timezone
-}
+$timezone = $module.Params.timezone
```
jborean93, line 20: Instead of using `Parse-Args` and `Get-AnsibleParam` we define the argument spec inside a hashtable, create the module object, then retrieve the module options through `$module.Params.x`. The `$spec` is the module option specification and this module only has 1 option called `timezone` that is required and is typed as a string.

We also explicitly have to opt into saying we support check mode with `supports_check_mode = $true` in the root of the spec hashtable. Without this the module will be skipped when running under check mode.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
 
-Try {
-    # Get the current timezone set
-    $result.previous_timezone = $(tzutil.exe /g)
-    If ($LASTEXITCODE -ne 0) {
-        Throw "An error occurred when getting the current machine's timezone setting."
-    }
+$module.Result.previous_timezone = $null
+$module.Result.timezone = $timezone
```
jborean93, line 23: The module result is located under `$module.Result`. We assign the return values accordingly.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
+        $Module.Result.stderr = $stderr -join "`n"
+        $Module.Result.rc = $LASTEXITCODE
+        $Module.FailJson("An error occurred when $Action.")
     }
+
+    $stdout
+}
+
+# Get the current timezone set
+$previousTz = Invoke-TzUtil /g -Module $module -Action "getting the current machine's timezone setting"
```
jborean93, line 55: The `Invoke-TzUtil` is merely a helper function and isn't strictly related to using the `Ansible.Basic` style. I just saw this module called `tzutil.exe` multiple times and was checking the result before failing if it wasn't `0`. So instead of duplicating those checks I put it in one function and have that handle all the error handling.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
-            }
-        }
+    $stdout = $null
+    $stderr = . { tzutil.exe @ArgumentList | Set-Variable -Name stdout } 2>&1 | ForEach-Object ToString
+
+    if ($LASTEXITCODE) {
+        $Module.Result.stdout = $stdout -join "`n"
+        $Module.Result.stderr = $stderr -join "`n"
+        $Module.Result.rc = $LASTEXITCODE
+        $Module.FailJson("An error occurred when $Action.")
```
jborean93, line 48: Instead of `Fail-Json $result "msg"` we now do `$module.FailJson("msg")`. The result is already stored in the module object so we just need to supply the error message to fail with.

`$module.FailJson()` will automatically exit the module and set it as failed.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
     }
+
+    $stdout
+}
+
+# Get the current timezone set
+$previousTz = Invoke-TzUtil /g -Module $module -Action "getting the current machine's timezone setting"
+$module.Result.previous_timezone = $previousTz
+
+if ($module.DiffMode) {
```
jborean93, line 58: We use `$module.DiffMode` to check if `--diff` is enabled for the task. This isn't strictly necessary but it avoids sending extra data back to Ansible.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
+    $module.Diff.before = "$previousTz`n"
+    $module.Diff.after = "$timezone`n"
```
jborean93, line 60: Instead of `diff` being part of the result, it's stored under a specific property `$module.Diff`. The structure is the same as before where we typically return a `before` and `after` with the changes involved.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
+
+if ($previousTz -ne $timezone) {
+    # Check that timezone is listed as an available timezone to the machine
+    $tzList = Invoke-TzUtil /l -Module $module -Action "listing the available timezones"
+
+    if ($tzList -notcontains ($timezone -replace '_dstoff')) {
+        $module.FailJson("The specified timezone: $timezone isn't supported on the machine.")
+    }
+
+    if (-not $module.CheckMode) {
```
jborean93, line 71: `$module.CheckMode` is a boolean that is true when we are running in check mode. Basically we want to avoid making any actual changes while in check mode.

plugins/modules/win_timezone.ps1
```diff
@@ -3,71 +3,75 @@
...
+        $null = Invoke-TzUtil /s $timezone -Module $module -Action "setting the specified timezone"
+    }
+    $module.Result.changed = $true
 }
 
-Exit-Json $result
+$module.ExitJson()
```
jborean93, line 77: Instead of `Exit-Json $result` we just do `$module.ExitJson()`. The result is already stored on the `$module` object so we don't have to provide it separately. This will stop the module and return the data in the result.

PR #718, opened 2024-12-30
win_get_url - Make checksum work more like ansible.builtin.get_url

### LINKED ISSUE #717: win_get_url checksum does not work as expected with force=false
##### SUMMARY

There is currently no way to use win_get_url and be sure that the expected file is in place without doing extra HTTP requests that waste resources.

If using a checksum and force=true (which to me is a surprising default) the file is always re-downloaded, which if it is pointed to public download sources is inefficient and consumes other peoples resources for no good reason. If force=false a HTTP HEAD request with If-Modified-Since is done, but this is not always enough to ensure that the correct file is downloaded (the local file might have a later modification time than the file on the server).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

win_get_url

##### ADDITIONAL INFORMATION

There are two ways in which the behaviour of the posix get_url module is better:

1. If the dest file exists and has the correct checksum no HTTP request is done. I cannot see a reason why doing a HTTP request in this case actually makes sense.
2. If dest exists and checksum is different from what is passed in it works as if force=true was passed. Which means you cannot end up in the situation that the module returns ok/no change while the file on disk has a different checksum than what is requested.

If win_get_url was changed to do these two things we could set force=false and be confident that the contents of dest is the file corresponding to the checksum.

```bash
$ echo hello > ~/public_html/tmp/hello.txt 
$ echo hei > ~/public_html/tmp/hei.txt 
$ sha1sum ~/public_html/tmp/{hello,hei}.txt 
f572d396fae9206628714fb2ce00f72e94f2258f  /home/kbj/public_html/tmp/hello.txt
b7a7bf03dcafd4d48001d6a2a6fd2ceaefa4cc1e  /home/kbj/public_html/tmp/hei.txt
```

```yaml
- hosts: somewhere.example.invalid
  become: no
  tasks:
    - win_get_url:
        url: https://otherplace.example.invalid/~kbj/tmp/hello.txt
        dest: C:/Windows/Temp/greeting.txt
        checksum: f572d396fae9206628714fb2ce00f72e94f2258f
        force: false
    - name: This does not actually make sure greeting.txt has the expected checksum
        win_get_url:
        url: https://otherplace.example.invalid/~kbj/tmp/hei.txt
        dest: C:/Windows/Temp/greeting.txt
        checksum: b7a7bf03dcafd4d48001d6a2a6fd2ceaefa4cc1e
        force: false
```

This is obviously a contrived example, but replace dest with SomeSoftware.zip and the URLs by download URLs for SomeSoftware-1.2.3.zip and SomeSoftware-1.3.0.zip and you get different behaviour on hosts depending on if they downloaded version 1.2.3 before or after 1.3.0 was released.

KBjorndal-VizRT:
##### SUMMARY

Makes checksum handling behave more like ansible.builtin.get_url

Fixes #717 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_get_url

##### ADDITIONAL INFORMATION

I'm not 100% sure about the logic to trigger collecting `$dest_checksum`, it is technically called in some circumstances it wasn't before, though if I read the code right that was only in error cases since all normal exits would calculate it at the end of the script.

### DIFF
--- a/changelogs/fragments/717-win_get_url-checksum.yml
+++ b/changelogs/fragments/717-win_get_url-checksum.yml
@@ -1,0 +1,7 @@
+minor_changes:
+  - win_get_url - if checksum is passed and destination file exists with
+    identical checksum no download is done unless force=yes
+    (https://github.com/ansible-collections/ansible.windows/issues/717)
+  - win_get_url - if checksum is passed and destination file exists with
+    different checksum file is always downloaded
+    (https://github.com/ansible-collections/ansible.windows/issues/717)
--- a/plugins/modules/win_get_url.ps1
+++ b/plugins/modules/win_get_url.ps1
@@ -151,6 +151,7 @@ Function Invoke-DownloadFile {
         [Parameter(Mandatory = $true)][Uri]$Uri,
         [Parameter(Mandatory = $true)][String]$Dest,
         [String]$Checksum,
+        [String]$DestChecksum,
         [String]$ChecksumAlgorithm
     )
 
@@ -187,14 +188,12 @@ Function Invoke-DownloadFile {
 
         $download = $true
         if (Test-Path -LiteralPath $Dest) {
-            # Validate the remote checksum against the existing downloaded file
-            $dest_checksum = Get-Checksum -Path $Dest -Algorithm $ChecksumAlgorithm
-
+            # Validate the remote checksum against the existing downloaded file.
             # If we don't need to download anything, save the dest checksum so we don't waste time calculating it
             # again at the end of the script
-            if ($dest_checksum -eq $tmp_checksum) {
+            if ($DestChecksum -eq $tmp_checksum) {
                 $download = $false
-                $Module.Result.checksum_dest = $dest_checksum
+                $Module.Result.checksum_dest = $DestChecksum
                 $Module.Result.size = (Get-AnsibleItem -Path $Dest).Length
             }
         }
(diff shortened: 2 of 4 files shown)


jborean93 (changes requested):
The changes you've added here look good. Appreciate you updating the docs and adding the changelog fragment but are you also able to add a test for this specific scenario. We currently have some tests with the `checksum` option under https://github.com/ansible-collections/ansible.windows/blob/main/tests/integration/targets/win_get_url/tasks/tests_checksum.yml but it doesn't look like there are any ones for idempotency checking when `force: false` is set.

If you need help with any of this please let me know and I can guide you where I can.

KBjorndal-VizRT:
> The changes you've added here look good. Appreciate you updating the docs and adding the changelog fragment but are you also able to add a test for this specific scenario. We currently have some tests with the `checksum` option under https://github.com/ansible-collections/ansible.windows/blob/main/tests/integration/targets/win_get_url/tasks/tests_checksum.yml but it doesn't look like there are any ones for idempotency checking when `force: false` is set.

I added a test for the behaviour where it shouldn't do any download if checksum already matches.

Tried a few different ways but can't think of how to make an automated test for the other part. As far as I can see httpbin only sends 304 replies for the /cache endpoint, which otherwise returns dynamic data so it's impossible to make it emulate the case where we have a file that differs by checksum but the If-Modified-Since logic would say there is nothing new at the URL.

KBjorndal-VizRT:
Not sure what the failing check here is about, is it related to the PR or some CI infrastructure?

jborean93:
> I added a test for the behaviour where it shouldn't do any download if checksum already matches.

Thanks, appreciate it.

> Tried a few different ways but can't think of how to make an automated test for the other part. As far as I can see httpbin only sends 304 replies for the /cache endpoint, which otherwise returns dynamic data so it's impossible to make it emulate the case where we have a file that differs by checksum but the If-Modified-Since logic would say there is nothing new at the URL.

I think that's fine to ignore for now, if we get any future regressions we can look at potentially spinning up a custom HTTP endpoint just as a test but I think that's overkill for now.

> Not sure what the failing check here is about, is it related to the PR or some CI infrastructure?

There seems to be some trouble with Zuul so the failure is unrelated here.

jborean93 (approved):

PR #719, opened 2025-01-02
Migrate win mapped drive

Yaish25491:
##### SUMMARY
Updating tests with facts instead of literal path.
Update Documentation.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
win_mapped_drive

##### ADDITIONAL INFORMATION
Allows you to modify mapped network drives for individual users.
Also support WebDAV endpoints in the UNC form.

### DIFF
--- a/plugins/modules/win_mapped_drive.py
+++ b/plugins/modules/win_mapped_drive.py
@@ -11,6 +11,11 @@ short_description: Map network drives for users
 description:
 - Allows you to modify mapped network drives for individual users.
 - Also support WebDAV endpoints in the UNC form.
+- While this module can manage mapped drives, you will not be able to use those
+  mapped drives in any subsequent Ansible task. Mapped drives are only visible
+  on interactive logons like through RDP and not network logons like WinRM or
+  SSH that Ansible uses to invoke tasks.
+version_added: 2.7.0
 options:
   letter:
     description:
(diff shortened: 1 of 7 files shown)


plugins/modules/win_mapped_drive.py
```diff
@@ -0,0 +1,148 @@
...
+# Copyright: (c) 2017, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_mapped_drive
+short_description: Map network drives for users
+description:
+- Allows you to modify mapped network drives for individual users.
+- Also support WebDAV endpoints in the UNC form.
```
jborean93, line 13: Some people think that they can use this module to map a drive for use in a subsequent Ansible task. I think we need to make it clearer that it is not the case and putting it in the description brings it right up to the top of the documentation for users. We also need to move the `version_added` entry to the root of the docs and not part of the `letter` option.

```suggestion
- Also support WebDAV endpoints in the UNC form.
- While this module can manage mapped drives, you will not be able to use those
  mapped drives in any subsequent Ansible task. Mapped drives are only visible
  on interactive logons like through RDP and not network logons like WinRM or
  SSH that Ansible uses to invoke tasks.
version_added: 2.7.0
```

jborean93 (changes requested):
Just a minor doc change is needed here.

plugins/modules/win_mapped_drive.py
Yaish25491: Changes applyed

jborean93 (approved):

jborean93:
Zuul seems to be having some troubles, going to ignore that failure for now.

PR #737, opened 2025-01-19
Migrate win http proxy

amitosw15:
##### SUMMARY
Migrate http proxy module

##### ISSUE TYPE
New Module Pull Request

##### COMPONENT NAME
win http proxy module


jborean93:
Test is failing because `win_inet_proxy` has not been migrated but we cannot migrate that module without this one. Will ignore for now and make sure CI still works once both modules have been merged.

jborean93 (approved):

PR #750, opened 2025-01-23
Migrate to newer AnsibleRequires format

jborean93:
##### SUMMARY
Migrate the module_util requirements to the newer #AnsibleRequires format. This new format makes it easier to debug locally as it won't affect and of the runtime checks PowerShell itself does for modules.

##### ISSUE TYPE
- Feature Pull Request


jborean93:
Looks like more changes need to happen in Ansible to achieve this migration :(

Issue #596, opened 2024-03-27
win_package: support checksum verification


mathstuf:
##### SUMMARY
`win_package` should support checksum verification of network downloads

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`ansible.windows.win_package`

##### ADDITIONAL INFORMATION
Verify that the downloaded file is the one expected before execution.

```yaml
- name: Install foobar
  ansible.windows.win_package:
    path: '{{ some_network_url }}'
    checksum: 'sha256sum:deadbeef…'
```

jborean93:
This seems like a simple thing to implement. I'm not sure whether to split the algorithm selection into another option or to have the prefix like you've specified here or not though.

mathstuf:
The Posix actions seem to use prefixes but Windows-specific actions seem to prefer split. I certainly prefer the former as it is one less variable to juggle, but consistency might be more valuable…

PR #759, opened 2025-02-18
win_package: checksum verify the package before installation

mathstuf:
##### SUMMARY
Note that this is only performed when installation is performed as file download does not occur in check mode, so URL sources cannot be verified in check mode anyways.

Fixes: #596

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`win_package`

### DIFF
--- a/plugins/modules/win_package.ps1
+++ b/plugins/modules/win_package.ps1
@@ -1282,6 +1282,8 @@ $spec = @{
         expected_return_code = @{ type = "list"; elements = "int"; default = @(0, 3010) }
         path = @{ type = "str" }
         chdir = @{ type = "path" }
+        checksum = @{ type = 'str' }
+        checksum_algorithm = @{ type = 'str'; default = 'sha1'; choices = @("md5", "sha1", "sha256", "sha384", "sha512") }
         product_id = @{ type = "str" }
         state = @{
             type = "str"
@@ -1310,6 +1312,8 @@ $arguments = $module.Params.arguments
 $expectedReturnCode = $module.Params.expected_return_code
 $path = $module.Params.path
 $chdir = $module.Params.chdir
+$checksum = $module.Params.checksum
+$checksum_algorithm = $module.Params.checksum_algorithm
 $productId = $module.Params.product_id
 $state = $module.Params.state
 $createsPath = $module.Params.creates_path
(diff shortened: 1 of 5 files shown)


plugins/modules/win_package.ps1
```diff
@@ -1383,6 +1412,16 @@ try {
             $path = $tempFile
         }
 
+        if ($checksum_algorithm -and $state -eq 'present' -and $path) {
+            $tmp_checksum = Get-Checksum -Path $path -Algorithm $checksum_algorithm
```
jborean93, line 1416: We can just use `Get-FileHash` here and not need the custom function

```suggestion
            $tmp_checksum = (Get-FileHash -LiteralPath $path -Algorithm $checksum_algorithm).Hash
```

plugins/modules/win_package.py
```diff
@@ -39,6 +39,24 @@
       or uninstalling a package.
     - This is only used for the C(msi), C(msp), and C(registry) providers.
     type: path
+  checksum:
+    description:
+      - If a I(checksum) is passed to this parameter, the digest of the
+        package will be calculated before executing it to verify that the
+        transfer completed successfully.
+    type: str
```
jborean93, line 47: Need to specify when the option was added

```suggestion
    type: str
    version_added: 2.8.0
```

plugins/modules/win_package.py
```diff
@@ -39,6 +39,24 @@
...
+  checksum_algorithm:
+    description:
+      - Specifies the hashing algorithm used when calculating the checksum of
+        the remote and destination file.
```
jborean93, line 51: I think this should be `of the path provided` or something along those lines.

plugins/modules/win_package.py
```diff
@@ -39,6 +39,24 @@
...
+    type: str
+    choices:
+      - md5
+      - sha1
+      - sha256
+      - sha384
+      - sha512
+    default: sha1
```
jborean93, line 59:
```suggestion
    default: sha1
    version_added: 2.8.0
```

plugins/modules/win_package.py
```diff
@@ -39,6 +39,24 @@
       or uninstalling a package.
     - This is only used for the C(msi), C(msp), and C(registry) providers.
     type: path
+  checksum:
+    description:
+      - If a I(checksum) is passed to this parameter, the digest of the
+        package will be calculated before executing it to verify that the
+        transfer completed successfully.
```
jborean93, line 46: This is slightly confusing, maybe change transfer completed successfully to something like

> verify that the path or downloaded file has the expected contents

jborean93:
Would you be able to add some integration tests here to test out

+ A checksum mismatch
+ A checksum match
+ The return value is as expected

jborean93 (approved):

jborean93:
Thanks for working on this!

PR #760, opened 2025-02-18
win_package: support `--check`

mathstuf:
##### SUMMARY
When the file is not downloaded by the module, it fails if the input does not exist. However, it may be populated by a `win_get_url` action which also does nothing with `--check`. Skip the assertion in this case.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_package

### DIFF
--- a/changelogs/fragments/win_package-local-file-check.yml
+++ b/changelogs/fragments/win_package-local-file-check.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - win_package - Support check mode with local file path sources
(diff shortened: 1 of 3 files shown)


jborean93 (changes requested):
Thanks for the fix here, is it possible to add an integration test for this scenario? Putting it in something like [msi_tests.yml](https://github.com/ansible-collections/ansible.windows/blob/main/tests/integration/targets/win_package/tasks/msi_tests.yml) is probably the easiest thing here.

jborean93 (approved):

jborean93:
Thanks again.

PR #766, opened 2025-03-14
add: recconect

3sky:
Trying to add re-connection feature.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Reconnection for powershell script execution - `modules/win_powershell`


jborean93:
We've spoken about this offline and the changes here won't affect the reconnection side of the Ansible connection plugins. That logic is all handled in the connection plugin itself which is in the ansible/ansible repository.

Issue #667, opened 2024-10-15
Win_Package uninstallation will exit without doing anything when path is pointing to a URL


yangxinyun:
##### SUMMARY

When win_package path argument is pointing to a URL and state argument is set to absent, this module will exit without doing uninstallation via downloading the URL specified software package and issue uninstall command as per the arguments
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_package
##### ANSIBLE VERSION

```paste below
ansible-core 2.11
```

##### COLLECTION VERSION

```paste below
Ansible.Windows version 2.5.0
```

##### CONFIGURATION

```paste below
NA
```

##### OS / ENVIRONMENT

RedHat Linux

##### STEPS TO REPRODUCE

```yaml
- name: Uninstall python 
  ansible.windows.win_package:
    path: https://www.python.org/ftp/python/3.11.0/python-3.11.0-amd64.exe
    arguments:
      - /uninstall
      - /quiet
     state: absent

```

##### EXPECTED RESULTS

python installed by https://www.python.org/ftp/python/3.11.0/python-3.11.0-amd64.exe is removed accordingly.

##### ACTUAL RESULTS

```paste below
nothing happened, the module exit without any change
```

jborean93:
Sorry I missed this issue, I was out on leave during this month and it seems to have slipped through the cracks when I got back.

In this case I think we probably should error as we cannot determine if the package is installed from just the executable and without a `product_id` is specified. Currently the code is most likely just going through the checks and as nothing matches any products it is exiting early as it hasn't found anything to uninstall. Even if it did download the exe it wouldn't be able to derive the `product_id` from the executable as there is no common mechanism to do so for executables.

Unfortunately the only thing you really have available to you here is to:

+ Move to using an msi - not possible if the package doesn't provide an msi
+ Specify the `product_id` so the module can see the package is either installed or not

If you wanted you can use a modified form of this PowerShell script to get the product ids at runtime

```powershell
'SOFTWARE', 'SOFTWARE\Wow6432Node' | ForEach-Object {
    $getParams = @{
        Path = "HKLM:\$_\Microsoft\Windows\CurrentVersion\Uninstall\*"
        Name = 'DisplayName'
        ErrorAction = 'SilentlyContinue'
    }
    Get-ItemProperty @getParams | Select-Object -Property @(
        @{ N = 'Name'; E = { $_.DisplayName } },
[...]

PR #783, opened 2025-05-01
win_package: Fail uninstall when no product id is provided

Akasurde:
##### SUMMARY

* When path is specified as an URL with state as absent,
  it is difficult to derive product id, which is required for
  uninstalling the package.
  Gracefully fail and let user know about this requirement.

Fixes: #667

##### ISSUE TYPE
- Bugfix Pull Request


jborean93 (approved):
Thanks for working on this.

Issue #656, opened 2024-09-09
win_dns_client should support DNS suffix


TiTi:
##### SUMMARY
I'm trying to add a dns suffix on a windows server machine which is NOT in a domain.
win_dns_client does not support that unfortunately

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
ansible.windows.win_dns_client

##### ADDITIONAL INFORMATION
Doable with Get-DnsClientGlobalSetting and Set-DnsClientGlobalSetting

https://learn.microsoft.com/en-us/powershell/module/dnsclient/get-dnsclientglobalsetting
https://learn.microsoft.com/en-us/powershell/module/dnsclient/set-dnsclientglobalsetting

PR #784, opened 2025-05-05
win_dns_client: Add support for suffixsearchlist

Akasurde:
##### SUMMARY

* Added support for DNS suffixsearchlist

Fixes: #656

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/plugins/modules/win_dns_client.ps1
+++ b/plugins/modules/win_dns_client.ps1
@@ -361,7 +361,7 @@ Function Set-DnsClientAddress {
 
 Function Set-DnsClientSuffixSearchList {
     Param(
-        [string] $suffix_search_list
+        [string[]] $suffix_search_list
     )
 
     Write-DebugLog ("Setting DNS Suffix Search List ({0})" -f [string[]]$suffix_search_list -join ", ")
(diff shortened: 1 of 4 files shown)


plugins/modules/win_dns_client.ps1
```diff
@@ -341,6 +359,22 @@ Function Set-DnsClientAddress {
     }
 }
 
+Function Set-DnsClientSuffixSearchList {
+    Param(
+        [string] $suffix_search_list
```
jborean93, line 364: This will coerce the string array to a string, should this be `[string[]]` instead?

plugins/modules/win_dns_client.ps1
```diff
@@ -379,6 +423,25 @@ Try {
...
+Catch {
+    $excep = $_
+
+    Write-DebugLog "Exception: $($excep | Out-String)"
+
+    Throw
+}
+
+Try {
+    if (-not (Test-DnsSuffixSearchList $suffix_search_list)) {
```
jborean93, line 436: Might be best to move this whole try block into the above one so we aren't repeating the catch multiple times.

plugins/modules/win_dns_client.ps1
```diff
@@ -379,6 +423,25 @@ Try {
...
+        $result.changed = $true
+        if (-not $check_mode) {
+            Set-DnsClientSuffixSearchList $suffix_search_list
+        }
+        else {
+            Write-DebugLog "Check mode, skipping"
```
jborean93, line 442: Would be nice to add something to this string saying `skipping setting suffix_search_list` or something along those lines.

jborean93 (changes requested):

jborean93 (approved):
Thanks for the changes.

TiTi:
Thanks for this.
However it looks impossible to only set the global dns suffixsearchlist, as parameters adapter_names & dns_servers are mandatory.

Setting an invalid adapter name to match nothing triggers an error:
```
fatal: [TEST1]: FAILED! => {
    "changed": false,
    "msg": "Unhandled exception while executing module: No MSFT_NetAdapter objects found with property 'Name' equal to 'zzzzzzzzzzzzzzzzzzzz'.  Verify the value of the property and retry."
}
```

In my case, this is challenging as:
- playbook is common to lots of machines having different adapter names that i don't really know for sure which one is the main adapter name
- dns servers are usually provided by DHCP.
The doc. says "An empty list will configure the adapter to use the DHCP-assigned values on connections where DHCP is enabled, **or disable DNS lookup on statically-configured connections**." So empty list would be ok .... but i might have statically configured connection on some machines, thus it would remove the dns lookup :/

=> I'm afraid setting up adapter_names & dns_servers might break dns configuration for some of my machines... when i'm only interested in setting global dns suffixsearchlist

PR #800, opened 2025-06-25
win_copy: fix symlink information

Akasurde:
##### SUMMARY

* symlinks should return list of dict containing src and dest

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/action/win_copy.py
tests/unit/plugins/action/test_win_copy.py


jborean93 (changes requested):
Thanks for the fix, can you write a changelog fragment for this bugfix please.

jborean93 (approved):

PR #803, opened 2025-07-04
usage of connect_timeout parameter to increase the connection timeout

theputta:
##### SUMMARY

 - Highlights a parameter `connect_timeout` not previously demonstrated. 
 - enhanced timeout settings is enabled 

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
connect_timeout

### DIFF
--- a/plugins/modules/win_wait_for.py
+++ b/plugins/modules/win_wait_for.py
@@ -125,7 +125,7 @@ EXAMPLES = r'''
     path: C:\temp\log.txt
     state: absent
 
-- name: Wait for port 8080 on 192.168.1.10 with a increased connect timeout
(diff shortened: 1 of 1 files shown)


plugins/modules/win_wait_for.py
```diff
@@ -125,6 +125,13 @@
     path: C:\temp\log.txt
     state: absent
 
+- name: Wait for port 8080 on 192.168.1.10 with a increased connect timeout
```
Akasurde:
```suggestion
- name: Wait for port 8080 on 192.168.1.10 with a increased connection timeout
```

theputta: @Akasurde 
- updated as per your recommendations

PR #810, opened 2025-07-11
Using fact_path in ansible.windows.setup module

theputta:
##### SUMMARY  
`ansible.windows.setup` module that demonstrates the use of the `fact_path` parameter which showcases how to retrieve PowerShell or JSON-based facts from `C:\CustomFacts`.

##### ISSUE TYPE  
- Docs Pull Request

##### COMPONENT NAME  
`ansible.windows.setup` module

##### ADDITIONAL INFORMATION  
This update addresses a gap in the documentation by showing how to use `fact_path`, which was previously undocumented in the examples section. This is particularly helpful for users managing their own local facts for automation use cases.

```yaml
- name: Gather Windows facts including custom local facts from C:\CustomFacts
  ansible.windows.setup:
    fact_path: 'C:\\CustomFacts'
```

### DIFF
--- a/plugins/modules/setup.yml
+++ b/plugins/modules/setup.yml
@@ -61,4 +61,4 @@ EXAMPLES: |
 
   - name: Gather Windows facts including custom local facts from C:\CustomFacts
     ansible.windows.setup:
-      fact_path: 'C:\\CustomFacts'
+      fact_path: C:\CustomFacts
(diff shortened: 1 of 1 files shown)


plugins/modules/setup.yml
```diff
@@ -58,3 +58,7 @@ EXAMPLES: |
   - name: Gather all facts with a custom timeout on Windows host
     ansible.windows.setup:
       gather_timeout: 30
+
+  - name: Gather Windows facts including custom local facts from C:\CustomFacts
+    ansible.windows.setup:
+      fact_path: 'C:\\CustomFacts'
```
jborean93: You only need to double up on `\` if you have it enclosed in double quotes. Personally I think the quoting isn't needed here so can just make this simpler.

```suggestion
      fact_path: C:\CustomFacts
```

PR #830, opened 2025-09-22
win_get_url - Fix force=no not doing HEAD request if checksum is not set

KBjorndal-VizRT:
##### SUMMARY
Fix a regression from #718 that broke the if not modified mechanism with force=no

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_get_url

##### ADDITIONAL INFORMATION
Looked into adding a test for this but I can't think of how it could be done without a custom http endpoint as the only difference in the module output is how long it takes.


changelogs/fragments/win_get_url-force-checksum.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - win_get_url - Fix force=no not doing HEAD request if cehcksum is not set
```
jborean93:
```suggestion
  - win_get_url - Fix force=no not doing HEAD request if checksum is not set
```

KBjorndal-VizRT: :facepalm: 

Fixed.

PR #841, opened 2025-10-29
Add DSC 3.0 support

### LINKED ISSUE #840: Support DSC 3.0
##### SUMMARY

Support [DSC 3.0](https://github.com/PowerShell/DSC).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`win_dsc3`

##### ADDITIONAL INFORMATION

DSC3's API is trivially mappable to how ansible works internally.  I have a local implementation which has been in production use for several months, which I am now working towards upstreaming.

The largest unknown atm is creating sufficient integration tests.  I've skimmed through how `win_dsc` is tested, and it looks quite involved.

(These examples are almost-verbatim snippets from my own playbooks)

```yaml
- name: Install Visual Studio Build Tools
  win_dsc3:
    resource: Microsoft.WinGet/Package
    properties:
      id: Microsoft.VisualStudio.2022.BuildTools
      source: winget
      useLatest: true

- name: Install Visual Studio components
  win_dsc3:
    resource: Microsoft.VisualStudio.DSC/VSComponents
    properties:
      productId: Microsoft.VisualStudio.Product.BuildTools
      channelId: VisualStudio.17.Release
      components:
        - Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64
        - Microsoft.VisualStudio.Component.VC.14.44.17.14.ATL
        - Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64
        - Microsoft.VisualStudio.Component.VC.14.38.17.8.ATL
        - Microsoft.VisualStudio.Component.Windows11SDK.22621
```

```yaml
- name: Configure WinGet for user
  win_dsc3:
    resource: Microsoft.WinGet/UserSettingsFile
    properties:
      settings:
        visual:
          progressBar: disabled
        interactivity:
          disable: true
```

yangskyboxlabs:
Add win_dsc3 module which interfaces with DSC3.  Installation of dsc.exe must be handled seprately.

##### SUMMARY

Fixes #840 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`win_dsc3`

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/dsc3.ps1
+++ b/plugins/modules/dsc3.ps1
@@ -82,7 +82,6 @@ if ($dscReturn.ExitCode -ne 0) {
         default { "Error: ExitCode=$_" }
     }
     $module.FailJson($reason)
-    $module.ExitJson()
 }
 
 $module.Result.result = ConvertFrom-Json $dscReturn.Stdout
@@ -108,7 +107,6 @@ switch ($module.Result.result.metadata."Microsoft.DSC".operation) {
     }
     default {
         $module.FailJson("Unexpected operation result of type '$_'")
-        $module.ExitJson()
     }
 }
 
--- a/plugins/modules/dsc3.py
+++ b/plugins/modules/dsc3.py
@@ -8,6 +8,7 @@ DOCUMENTATION = r"""
 ---
 module: dsc3
 short_description: Sets or checks DSC v3 configuration state
+version_added: '3.4.0'
 description:
     - Calls C(dsc config set) or C(dsc config test) using O(config) as the configuration document.
     - This module assumes that C(dsc) can be found using E(PATH) environment variable, and C(dsc)


plugins/modules/win_dsc3.py
```diff
@@ -0,0 +1,179 @@
...
+      - --id=Microsoft.DSC
+      - --exact
+      - --source=winget
+      - --scope=machine
+      - --accept-package-agreements
+      - --accept-source-agreements
+      - --disable-interactivity
+    creates: '{{ ansible_env.ProgramFiles }}\WinGet\Links\dsc.exe'
+
+- name: Install .NET Framework SDK form winget
```
yangskyboxlabs:
```suggestion
- name: Install .NET Framework SDK from winget
```

yangskyboxlabs:
Still on my radar, but may not make meaningful changes to the PR until 2026.

jborean93:
Take all the time you need. Appreciate you putting in the work here.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+
+$traceLevel = switch ($module.Verbosity) {
+    3 { "info" }
+    4 { "debug" }
+    5 { "trace" }
+    default { "warn" }
+}
+
+$inputJson = $configDoc | ConvertTo-Json -Depth 100 -Compress
+if ($module.Verbosity -ge 4) {
```
yangskyboxlabs: This might need to be documented, or removed

tests/integration/targets/dsc3/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+win_dsc3_zip_url: https://github.com/PowerShell/DSC/releases/download/v3.1.2/DSC-3.1.2-x86_64-pc-windows-msvc.zip
```
yangskyboxlabs, line 2: Looking at other integration tests, things like this are mirrored.

It's probably OK to do the same here.

plugins/modules/win_dsc3.py
```diff
@@ -0,0 +1,189 @@
...
+    - Calls C(dsc config set), C(dsc config test), or C(dsc config get) using the task definition
+      as the configuration document.
+    - This task expects the same schema for a DSC configuration document, but specific required
+      properties are assigned appropriate default values when omitted. See
+      U(https://learn.microsoft.com/en-us/powershell/dsc/reference/schemas/config/document) for the
+      full specification.
+author:
+    - Yang Zhao (@yangskyboxlabs)
+options:
+    resources:
```
yangskyboxlabs: I see several indentation styles used for docs. Not sure what is the most preferred.

I've arbitrarily chosen 4 spaces here.

tests/integration/targets/dsc3/files/fake/fake.resource.ps1
```diff
@@ -0,0 +1,45 @@
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
```
yangskyboxlabs, line 1: This fake implementation relies on `pwsh` to get consistent `utf-8` and JSON handling. I don't think I managed to make it work under `powershell`, although that was several months ago.

If this is potentially a problem, we'll need to come with something else.

plugins/modules/win_dsc3.py
```diff
@@ -0,0 +1,189 @@
...
+        type: dict
+
+    command:
+        description:
+            - DSC config command to use.
+            - When O(command=set) and C(check_mode=True), C(test) is used for final invocation.
+        type: str
+        choices: [ set, get ]
+        default: set
+    config_options:
```
yangskyboxlabs: Thinking about it more, it may make more sense to rename this something like `extra_options`, and move `raw_results`, `extra_paths`, and `resource_paths` here.  That would more clearly delineate what properties are mirrored from DSC config docs, and what are "runtime options".

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+        parameters = @{ type = "dict" }
+        variables = @{ type = "dict" }
+        resources = @{
+            type = "list"
+            elements = "dict"
+            options = @{
+                type = @{ required = $true }
+                name = @{ required = $true }
+                properties = @{ type = "dict"; required = $true }
+                dependsOn = @{ type = "list"; elements = "str" }
```
jborean93: We try and conform to the `snake_case` naming scheme. I assume DSC uses `dependsOn` but is it going to be a problem if we rename this to `depends_on`?

plugins/modules/win_dsc3.py
jborean93: That's totally fine, key thing is that it is consistent in that doc.

plugins/modules/win_dsc3.py
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: Contributors to the Ansible project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+---
+module: win_dsc3
```
jborean93: What are your thoughts on dropping the `win_` prefix and just making this `dsc3`?

I know all the other modules are `win_` but in some of the new collections for Windows I've mostly dropped this in favour of specifying the FQ name. Doing `ansible.windows.dsc3` is going to be shorter than `ansible.windows.win_dsc3` and the `windows` in the collection name somewhat negates the `win_` prefix anyway.

Theoretically with the PowerShell 7 work I've got in the pipeline this may even be able to run on non-Windows as well.

The downside is that it goes against the established norm in this collection.

plugins/modules/win_dsc3.py
```diff
@@ -0,0 +1,189 @@
...
+- name: Install .NET Framework SDK from winget
+  ansible.windows.win_dsc3:
+    resources:
+      - name: Install .NET Framework
+        type: Microsoft.WinGet/Package
+        properties:
+          id: Microsoft.DotNet.Framework.DeveloperPack_4
+          source: winget
+
+# Manage VisualStudio BuildTools using its DSC2 resource
```
jborean93: I'm a bit reluctant on documenting this right now. I'm hoping to land native PowerShell 7 support outside of PSRP and it uses a different set of options. Maybe let's just keep this off the examples for now.

plugins/modules/win_dsc3.ps1
jborean93: Might be best to remove for now. We can always revisit this if we find that knowing this value is useful. It doesn't seem too useful considering the `$configDoc` is essentially the input values with `-vvv` will show anyway.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+    variables = $module.Params.variables
+    resources = $module.Params.resources
+}
+
+$traceLevel = switch ($module.Verbosity) {
+    3 { "info" }
+    4 { "debug" }
+    5 { "trace" }
+    default { "warn" }
+}
```
jborean93: I know `win_dsc` does this (a mistake on my part) but I think it'll be better to just have this as part of the module options. That way users don't need to enable higher verbosity levels and sift through the extra output just to get this information.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+        $paramValuesArgs
+        $command,
+        "--file=-"
+        "--output-format=json"
+    )
+    InputObject = $inputJson
+}
+
+# Add entries to PATH, if requested
+foreach ($p in $module.Params.extra_paths) {
```
jborean93: I don't think this should be a module option. We have the `environment` keyword which can manipulate the `PATH` env var for a module if that's required.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+    $Env:PATH += ";$p"
+}
+# Ensure winget installed instance is always found, even if PATH has not been updated
+$Env:PATH += ";$Env:ProgramFiles\WinGet\Links\;$Env:LOCALAPPDATA\Microsoft\WinGet\Links\"
```
jborean93: Why is this required, does `dsc` require this, when would they have not been set?

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+
+# Set DSC_RESOURCE_PATH, if requested
+if ($module.Params.resource_paths) {
```
jborean93: Same here, I think we should just leave this to the `environment` directive in the task.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+    $Env:DSC_RESOURCE_PATH = $module.Params.resource_paths -join ";"
+}
+
+$dscReturn = Start-AnsibleWindowsProcess @dscArgs
+$module.Result.rc = $dscReturn.ExitCode
+$module.Result.stderr = $dscReturn.Stderr
+
+if ($dscReturn.ExitCode -ne 0) {
```
jborean93: Does DSC have some return code for a reboot is required or is it just 0 is success and everything else is not?

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+$Env:PATH += ";$Env:ProgramFiles\WinGet\Links\;$Env:LOCALAPPDATA\Microsoft\WinGet\Links\"
+
+# Set DSC_RESOURCE_PATH, if requested
+if ($module.Params.resource_paths) {
+    $Env:DSC_RESOURCE_PATH = $module.Params.resource_paths -join ";"
+}
+
+$dscReturn = Start-AnsibleWindowsProcess @dscArgs
+$module.Result.rc = $dscReturn.ExitCode
+$module.Result.stderr = $dscReturn.Stderr
```
jborean93: Will `Stderr` only ever contain values when `ExitCode -ne 0`? If so I think we should just set this in the if block below on a failure rather than always returning it. Same for `$module.Result.rc` as if it was successful then `0` is assumed. They seem to only be useful on failures for debugging purposes.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        command = @{ choices = ( "set", "get" ); default = "set" }
+        config_options = @{
+            type = "dict"
+            options = @{
+                parameters = @{ type = "dict" }
+            }
+        }
+        raw_results = @{ type = "bool"; default = $false }
```
jborean93: Would it just be better and simpler to always return the raw results. That way we aren't getting in the middle of anything and aren't omitting information that might be useful for end users in that result. What benefit do we get from trying to create our own result structure?

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,192 @@
...
+        resource_paths = @{ type = "list"; elements = "path" }
+    }
+    supports_check_mode = $true
+}
+
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$command = $module.Params.command
+if (($command -eq "set") -and ($module.CheckMode)) {
+    $command = "test"
```
jborean93: Is the test output any different structure wise to `set`? Just trying to see whether someone relying on some output value might then break if done in check mode.

tests/integration/targets/dsc3/defaults/main.yml
jborean93: Considering we are running from GH if it goes down then we probably have more issues. I can certainly mirror it if need be in the future.

tests/integration/targets/win_dsc3/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+        dest: '{{ test_dsc3_local_dir }}'
+        mode: '0666'
+      delegate_to: localhost
+      connection: local
+      when: (test_dsc3_local_dir + "/dsc.exe") is not exists
+
+    - name: Install DSC to remote temp dir
+      ansible.windows.win_copy:
+        src: files/dsc/
+        dest: '{{ test_dsc3_path }}'
```
jborean93: Can we instead use `win_get_url` to download the archive into the temp dir on the Windows path and maybe just a `win_powershell` step to extract that if the dest exe isn't present? That way we don't have to worry about cleaning up anything locally and avoid the hit of `win_copy` from having to copy all the files in there.

tests/integration/targets/win_dsc3/tasks/tests.yml
```diff
@@ -0,0 +1,249 @@
...
+          - name: Echo parameter
+            type: Microsoft.DSC.Debug/Echo
+            properties:
+              output: "[parameters('message')]"
+        parameters:
+          message:
+            type: string
+        config_options:
+          parameters:
+            message: Hello, world!
```
jborean93: It might be nice to stress test this a bit more and embed a JSON string instead like `message: Hello, {"location": "world"}`. That way we are ensuring that we are correctly escaping the characters needed when calling `dsc.exe` as the internal JSON string will contain a combination of `"` and `\"`.

plugins/modules/win_dsc3.ps1
jborean93: I forgot to mention, if the goal was to provide a custom location to `dsc.exe` that wasn't on the path we typically just expose that as an option like `dsc_executable` that has a default of `dsc.exe` which would lookup the `PATH` location.

tests/integration/targets/win_dsc3/tasks/tests.yml
yangskyboxlabs: That's a good idea.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: They each have their own separate formal schema, but appear to be identical in practice.  At least for now.

yangskyboxlabs: set: https://learn.microsoft.com/en-us/powershell/dsc/reference/schemas/outputs/config/set?view=dsc-3.0
test: https://learn.microsoft.com/en-us/powershell/dsc/reference/schemas/outputs/config/test?view=dsc-3.0

plugins/modules/win_dsc3.ps1
yangskyboxlabs: These fields are part of the DSC resource instance schema (https://learn.microsoft.com/en-us/powershell/dsc/reference/schemas/config/resource?view=dsc-3.0), so using `snake_case` would automatically result in validation failures.  Since it's an optional property, I think it's fine to just leave it out of the module schema.

plugins/modules/win_dsc3.py
yangskyboxlabs: 👍  Will remove.

plugins/modules/win_dsc3.py
yangskyboxlabs: I would love to drop the `win_` prefix.  I assume it made a lot of sense before FQ usage became the recommended default.

I can do a rename pass at the end once we've sorted out the other implementation details.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: I think I added this mostly to make the integration tests work. For some reason, it just slipped my mind that I can manipulate `PATH` directly.  Will remove.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: Will remove.

It's something carried over from a very early version that removed a bit of friction from my particular deployment context.  I agree the module shouldn't be doing this internally.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: Each resource can flag a reboot request as part of its `set` operation: https://learn.microsoft.com/en-us/powershell/dsc/reference/schemas/resource/properties/rebootrequested?view=dsc-3.0

I was actually looking to hook this feature up to something, but didn't immediately find _where_ to connect it to.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: All logging messages go to `stderr`, so capturing it unconditionally makes warnings visible, and removes the need to special case when diagnostic logging is enabled.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: 👍  This felt wrong to me as well. It would be cleaner as its own option.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: Two main reasons:
- The raw output is quite large. It includes full copies of the "before/actual" and "after/desired" state tree, as well as metadata with several timestamps. It's quite difficult to determine the current state at-a-glance when looking at Ansible's output.
- `set` and `test` actually return documents with different schema, with completely different property names for required data.  This difference when using check mode is fairly easy to map into "Ansible semantics" at the module level, but can be very complicated to handle for downstream tasks as-is.

For generating `before`/`after` diffs, there's also a UX issue with raw results since DSC provides them in different properties.  We would need to do something very clever to make "diffing" the raw results usable, and I'm not sure that we can.

For comparison, this is the diff output that's currently produced using the synthetic `config_results` property:
```
TASK [win_dsc3 : Configure new state] ******************************************
--- before
+++ after
@@ -2,7 +2,10 @@
     "config_results": [
         {
             "name": "New state",
-            "state": {}
+            "state": {
+                "number": 1,
+                "string": "text"
+            }
         }
     ]
 }
```

yangskyboxlabs: And, for further context, this is the raw DSC output on my machine when used with `Microsoft.VisualStudio.DSC/VSComponents`:

```
PS D:\ansible> dsc --trace-level=error config test -f .\vs.dsc.yml
metadata:
  Microsoft.DSC:
    version: 3.1.0
    operation: test
    executionType: actual
    startDatetime: 2026-01-16T11:13:49.693646700-08:00
    endDatetime: 2026-01-16T11:13:59.267836-08:00
    duration: PT9.5741893S
    securityContext: elevated
results:
- metadata:
    Microsoft.DSC:
      duration: PT4.7271914S
  name: Visual Studio Components
  type: Microsoft.VisualStudio.DSC/VSComponents
  result:
    desiredState:
      productId: Microsoft.VisualStudio.Product.Professional
      channelId: VisualStudio.17.Release
      components:
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ATL
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.ATL
      - Microsoft.VisualStudio.Component.Windows11SDK.22621
    actualState:
      InDesiredState: true
    inDesiredState: true
    differingProperties:
    - productId
    - channelId
    - components
messages: []
hadErrors: false
```

```
PS D:\ansible> dsc --trace-level=error config set -f .\vs.dsc.yml
metadata:
  Microsoft.DSC:
    version: 3.1.0
    operation: set
    executionType: actual
    startDatetime: 2026-01-16T11:14:14.193701600-08:00
    endDatetime: 2026-01-16T11:14:25.452596700-08:00
    duration: PT11.2588951S
    securityContext: elevated
results:
- metadata:
    Microsoft.DSC:
      duration: PT6.5613582S
  name: Visual Studio Components
  type: Microsoft.VisualStudio.DSC/VSComponents
  result:
    beforeState:
      components:
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ATL
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.ATL
      - Microsoft.VisualStudio.Component.Windows11SDK.22621
      includeOptional: false
      productId: Microsoft.VisualStudio.Product.Professional
      includeRecommended: false
      vsConfigFile: ''
      installedComponents:
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ARM64
      - Microsoft.VisualStudio.Component.VC.Tools.ARM64EC
      - Microsoft.VisualStudio.Component.VC.Llvm.Clang
      - Microsoft.VisualStudio.Component.FSharp
      - Component.Microsoft.Windows.DriverKit
      - Microsoft.VisualStudio.Component.Windows11SDK.26100
      - Microsoft.NetCore.Component.Runtime.8.0
      - Microsoft.NetCore.Component.Runtime.9.0
      - Microsoft.VisualStudio.Component.Windows10SDK
      - Microsoft.VisualStudio.Component.VC.CoreIde
      - Microsoft.ComponentGroup.ClickOnce.Publish
      - Microsoft.VisualStudio.Component.Debugger.JustInTime
      - Microsoft.Net.Component.4.6.2.TargetingPack
      - Microsoft.VisualStudio.Component.EntityFramework
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ATL
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64
      - Microsoft.ComponentGroup.Blend
      - Microsoft.VisualStudio.Component.DotNetModelBuilder
      - Microsoft.VisualStudio.Component.ManagedDesktop.Prerequisites
      - Microsoft.VisualStudio.Component.VC.14.42.17.12.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.ATL
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64
      - Microsoft.VisualStudio.Component.IntelliCode
      - Microsoft.VisualStudio.Component.Unreal.TestAdapter
      - Microsoft.VisualStudio.Component.VC.DiagnosticTools
      - Microsoft.VisualStudio.Component.Graphics.Tools
      - Component.Unreal.Ide
      - Microsoft.VisualStudio.Component.VC.Tools.x86.x64
      - Microsoft.VisualStudio.Component.NuGet.BuildTools
      - Microsoft.Net.ComponentGroup.TargetingPacks.Common
      - Microsoft.VisualStudio.Workload.NativeDesktop
      - Microsoft.VisualStudio.Component.Windows11SDK.22621
      - Microsoft.VisualStudio.Component.Vcpkg
      - Component.Unreal.Debugger
      - Microsoft.VisualStudio.Component.DiagnosticTools
      - Microsoft.VisualStudio.Component.VC.ASAN
      - Microsoft.VisualStudio.Component.ManagedDesktop.Core
      - Microsoft.Component.ClickOnce
      - Microsoft.VisualStudio.Component.SQL.CLR
      - Microsoft.VisualStudio.Component.NuGet
      - Microsoft.VisualStudio.Component.TextTemplating
      - Microsoft.VisualStudio.Component.Roslyn.LanguageServices
      - Microsoft.VisualStudio.Component.TypeScript.TSServer
      - Microsoft.Net.ComponentGroup.DevelopmentPrerequisites
      - Component.Microsoft.VisualStudio.LiveShare.2022
      - Microsoft.NetCore.Component.SDK
      - Microsoft.Net.ComponentGroup.4.8.DeveloperTools
      - Microsoft.Net.Component.4.8.TargetingPack
      - Component.VisualStudio.GitHub.Copilot
      - Microsoft.VisualStudio.Component.VC.TestAdapterForGoogleTest
      - Microsoft.VisualStudio.Component.Windows10SDK.19041
      - Microsoft.VisualStudio.Component.VC.TestAdapterForBoostTest
      - Microsoft.VisualStudio.Component.VC.CMake.Project
      - Microsoft.VisualStudio.ComponentGroup.WebToolsExtensions.CMake
      - Microsoft.VisualStudio.Component.CppBuildInsights
      - Microsoft.NetCore.Component.DevelopmentTools
      - Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Core
      - Microsoft.VisualStudio.Component.VC.Redist.14.Latest
      - Microsoft.VisualStudio.Component.VC.ATL
      - Microsoft.VisualStudio.Workload.ManagedDesktop
      - Microsoft.Component.MSBuild
      - Microsoft.VisualStudio.Component.Roslyn.Compiler
      - Microsoft.VisualStudio.Component.JavaScript.TypeScript
      - Microsoft.VisualStudio.ComponentGroup.WebToolsExtensions
      - Microsoft.Net.Component.4.7.2.TargetingPack
      - Microsoft.Net.Component.4.8.SDK
      - Microsoft.VisualStudio.Workload.CoreEditor
      - Microsoft.VisualStudio.Component.CoreEditor
      allowUnsignedExtensions: false
      channelId: VisualStudio.17.Release
    afterState:
      channelId: VisualStudio.17.Release
      includeRecommended: false
      allowUnsignedExtensions: false
      vsConfigFile: ''
      productId: Microsoft.VisualStudio.Product.Professional
      includeOptional: false
      components:
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ATL
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.ATL
      - Microsoft.VisualStudio.Component.Windows11SDK.22621
      installedComponents:
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ARM64
      - Microsoft.VisualStudio.Component.VC.Tools.ARM64EC
      - Microsoft.VisualStudio.Component.VC.Llvm.Clang
      - Microsoft.VisualStudio.Component.FSharp
      - Component.Microsoft.Windows.DriverKit
      - Microsoft.VisualStudio.Component.Windows11SDK.26100
      - Microsoft.NetCore.Component.Runtime.8.0
      - Microsoft.NetCore.Component.Runtime.9.0
      - Microsoft.VisualStudio.Component.Windows10SDK
      - Microsoft.VisualStudio.Component.VC.CoreIde
      - Microsoft.ComponentGroup.ClickOnce.Publish
      - Microsoft.VisualStudio.Component.Debugger.JustInTime
      - Microsoft.Net.Component.4.6.2.TargetingPack
      - Microsoft.VisualStudio.Component.EntityFramework
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.ATL
      - Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64
      - Microsoft.ComponentGroup.Blend
      - Microsoft.VisualStudio.Component.DotNetModelBuilder
      - Microsoft.VisualStudio.Component.ManagedDesktop.Prerequisites
      - Microsoft.VisualStudio.Component.VC.14.42.17.12.x86.x64
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.ATL
      - Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64
      - Microsoft.VisualStudio.Component.IntelliCode
      - Microsoft.VisualStudio.Component.Unreal.TestAdapter
      - Microsoft.VisualStudio.Component.VC.DiagnosticTools
      - Microsoft.VisualStudio.Component.Graphics.Tools
      - Component.Unreal.Ide
      - Microsoft.VisualStudio.Component.VC.Tools.x86.x64
      - Microsoft.VisualStudio.Component.NuGet.BuildTools
      - Microsoft.Net.ComponentGroup.TargetingPacks.Common
      - Microsoft.VisualStudio.Workload.NativeDesktop
      - Microsoft.VisualStudio.Component.Windows11SDK.22621
      - Microsoft.VisualStudio.Component.Vcpkg
      - Component.Unreal.Debugger
      - Microsoft.VisualStudio.Component.DiagnosticTools
      - Microsoft.VisualStudio.Component.VC.ASAN
      - Microsoft.VisualStudio.Component.ManagedDesktop.Core
      - Microsoft.Component.ClickOnce
      - Microsoft.VisualStudio.Component.SQL.CLR
      - Microsoft.VisualStudio.Component.NuGet
      - Microsoft.VisualStudio.Component.TextTemplating
      - Microsoft.VisualStudio.Component.Roslyn.LanguageServices
      - Microsoft.VisualStudio.Component.TypeScript.TSServer
      - Microsoft.Net.ComponentGroup.DevelopmentPrerequisites
      - Component.Microsoft.VisualStudio.LiveShare.2022
      - Microsoft.NetCore.Component.SDK
      - Microsoft.Net.ComponentGroup.4.8.DeveloperTools
      - Microsoft.Net.Component.4.8.TargetingPack
      - Component.VisualStudio.GitHub.Copilot
      - Microsoft.VisualStudio.Component.VC.TestAdapterForGoogleTest
      - Microsoft.VisualStudio.Component.Windows10SDK.19041
      - Microsoft.VisualStudio.Component.VC.TestAdapterForBoostTest
      - Microsoft.VisualStudio.Component.VC.CMake.Project
      - Microsoft.VisualStudio.ComponentGroup.WebToolsExtensions.CMake
      - Microsoft.VisualStudio.Component.CppBuildInsights
      - Microsoft.NetCore.Component.DevelopmentTools
      - Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Core
      - Microsoft.VisualStudio.Component.VC.Redist.14.Latest
      - Microsoft.VisualStudio.Component.VC.ATL
      - Microsoft.VisualStudio.Workload.ManagedDesktop
      - Microsoft.Component.MSBuild
      - Microsoft.VisualStudio.Component.Roslyn.Compiler
      - Microsoft.VisualStudio.Component.JavaScript.TypeScript
      - Microsoft.VisualStudio.ComponentGroup.WebToolsExtensions
      - Microsoft.Net.Component.4.7.2.TargetingPack
      - Microsoft.Net.Component.4.8.SDK
      - Microsoft.VisualStudio.Workload.CoreEditor
      - Microsoft.VisualStudio.Component.CoreEditor
    changedProperties: null
messages: []
hadErrors: false
```

plugins/modules/win_dsc3.ps1
yangskyboxlabs: Oh, just to clarify, they have the same structure _at the top level_.

The schema for the embedded resource states actually differ. See the thread re: `raw_results`.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: For note, `dsc` itself relies on `PATH` to discover resource implementations, so providing something like `dsc_executable` can potentially lead to unintuitive breakages.

I'll add a note in the documentation that people should manipulate `PATH` directly if `dsc` can't be found without it.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,156 @@
...
+        $module.Result.changed = $changed
+
+        $result
+    }
+)
+
+if ($stateDiffs) {
+    # Combine diffs from each resource so it matches the structure of 'config_results'
+    $diffs = @{
+        before = @{ config_results = @() }
```
yangskyboxlabs: Hm. Maybe drop the extra `config_results` layer, and just emit the diff for each resource state?

yangskyboxlabs: The default diff formatter expects `before` and `after` to be `dict`s, so flattening to a list doesn't accomplish what I want.

Will probably use `resources` instead of `config_results` to better match the input configuration.

plugins/modules/win_dsc3.py
jborean93: Sounds good!

plugins/modules/win_dsc3.ps1
jborean93: Typically a module returns `reboot_required` when a reboot was detected. It's not a hard and fast rule but it's common convention.

plugins/modules/win_dsc3.ps1
jborean93: Ah ok if `stderr` can contain extra info that's not error related then returning it always is ok.

plugins/modules/win_dsc3.ps1
jborean93: The complex error definitely still leans me towards not worrying about it and just returning the raw results as is. We can certainly hook in the extra work to make the diff prettier and return easier to access info like the `reboot_required` return values at the root level. My concerns with converting more things would be if a schema can change the output format which we will now need to handle so the less we interact with the better.

plugins/modules/win_dsc3.ps1
jborean93: We could make the before/after each have a list with each resource at that list's index. I'm pretty sure Ansible's diff handler will be able to deal with that.

plugins/modules/win_dsc3.ps1
```diff
@@ -0,0 +1,156 @@
...
+#AnsibleRequires -PowerShell ansible_collections.ansible.windows.plugins.module_utils.Process
+
+$spec = @{
+    options = @{
+        config = @{
+            type = "dict"
+            required = $true
+        }
+
+        config_options = @{
```
jborean93: Is there a good reason to have these options inside `config_options`. We could just put `parameters` and `trace_level` at the root level. Now that we have `config` there shouldn't be any conflicts with things that DSC would accept.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: I think the documentation on flagging reboot is now out of date, and the implementation in `dsc` itself is incomplete: https://github.com/PowerShell/DSC/pull/1260

It looks like this is going to be properly supported from `3.2.0` onwards, which is not yet released.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: I did a pass to simplify the diff generation, and that rewrite has convinced me it is indeed preferable to always return the raw results.

The new return now always populates `result` with the raw output object, following DSC's nomenclature used in its schemas.  The synthetic `state` property has been dropped entirely.

tests/integration/targets/win_dsc3/tasks/main.yml
yangskyboxlabs: Reworked these steps so it now does just that.

I've moved the target install path into `defaults/main.yml` so it can be overridden using `vars/main.yml` for local development.

plugins/modules/win_dsc3.ps1
yangskyboxlabs: The diff formatter doesn't seem to like arrays.

I've changed the diff layout to be like this, to better mirror what's actually defined in `config`:
```yaml
diff:
  before:
    resources:
      - name: ...
        type: ...
        properties:
          ...
  after:
    resources:
      - name: ...
        type: ...
        properties:
          ...
```

```diff
TASK [win_dsc3 : Check mode with changes] **************************************
--- before
+++ after
@@ -3,7 +3,7 @@
         {
             "name": "Check mode",
             "properties": {
-                "b": null
+                "b": "b"
             },
             "type": "Ansible.Windows.Debug/Fake"
         }
```

plugins/modules/win_dsc3.ps1
yangskyboxlabs: Yeah, this nesting doesn't make much sense after the parameter reorg.

I've bumped all the options up to top level again.

yangskyboxlabs:
I've renamed the module to `dsc3` but there appear to be Windows-specific logic in `ansible-test` that fails to include test modules that do not have some prefix. So the test target has been left as `win_dsc3`.

tests/integration/targets/dsc3/aliases
```diff
@@ -0,0 +1 @@
+windows
```
jborean93, line 1: We need to associate the test with a specific group. Looking at the overnight run, 3 seems to be the best choice here.

```suggestion
windows
shippable/windows/group3
```

jborean93 (approved):

jborean93:
You’ll need to set `[Console]::OutputEncoding = [Textt.UTF8Encoding]::new()` as the very first step. That should be enough to ensure the output strings are encoded as utf-8.

jborean93:
Awesome I’ll have one last look over on Monday before merging but there shouldn’t be anything more left to do. Thanks for all your hard work here!

plugins/modules/dsc3.ps1
```diff
@@ -0,0 +1,167 @@
...
+    $reason = switch ($dscReturn.ExitCode) {
+        1 { "Invalid arguments (ExitCode=1)" }
+        2 { "Resource error (ExitCode=2)" }
+        3 { "JSON serizliation error (ExitCode=3)" }
+        4 { "Input YAML or JSON is invalid (ExitCode=4)" }
+        5 { "Data failed schema validation (ExitCode=5)" }
+        default { "Error: ExitCode=$_" }
+    }
+    $module.FailJson($reason)
+    $module.ExitJson()
```
jborean93:
```suggestion
```

Technically `FailJson` exits as well so this will never be hit.

plugins/modules/dsc3.ps1
```diff
@@ -0,0 +1,167 @@
...
+        $changedPropertiesProp = "changedProperties"
+    }
+    "test" {
+        $beforeProp = "actualState"
+        $afterProp = "desiredState"
+        $changedPropertiesProp = "differingProperties"
+    }
+    default {
+        $module.FailJson("Unexpected operation result of type '$_'")
+        $module.ExitJson()
```
jborean93:
```suggestion
```

plugins/modules/dsc3.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: Contributors to the Ansible project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+---
+module: dsc3
+short_description: Sets or checks DSC v3 configuration state
```
jborean93, line 10:
```suggestion
short_description: Sets or checks DSC v3 configuration state
version_added: '3.4.0'

jborean93 (approved):
Just 3 minor changes that I'll commit here through suggestions but everything else is great.

PR #855, opened 2025-12-26
Feature: win_path prepend option

### LINKED ISSUE #661: add "prepend" option to win_path
##### SUMMARY
Add an option to the module that will prepend the specified path(s) to the existing path.  If the path(s) already exists, then move it(them) to the top.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
ansible.windows.win_path

##### ADDITIONAL INFORMATION

In some cases, we need a path element to occur before System32.  The existing code moves System32 later in the path, which could unintentionally cause executables in system32 to be overridden by other path elements.

```yaml
- name: Ensure that updated OpenSSH executables are preferred over older ones in System32
  ansible.windows.win_path:
    elements: 
      - 'C:\Program Files\OpenSSH'
      - '%SystemRoot%\system32'
    state: prepend
```

bishalprasad321:
##### SUMMARY

This PR introduces a new `prepend` parameter to the `win_path` module. The `prepend` parameter allows users to add specified path(s) to the top of the specified environment variable. If the paths already exist, they are moved to the top. This feature addresses the need to prioritize certain paths, such as ensuring executables in specific directories are found before those in `System32`.

Fixes #661 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`win_path`

##### ADDITIONAL INFORMATION

This change simplifies the interface by introducing a `prepend` boolean parameter instead of overloading the `state` parameter. The `prepend` parameter defaults to `false`, ensuring backward compatibility.

Prepend semantics are positional: elements are enforced as an ordered block at the start of `<env-var>`, and the module exits early when that invariant already holds.

Integration tests have been added to verify the following scenarios:
1. Prepending a single path.
2. Prepending multiple paths and verifying their order.
3. Idempotence of the `prepend` operation.
4. Moving existing paths to the top.
5. Check mode functionality for `prepend`.

Example usage:

```yaml
- name: Prepend a path to the PATH variable
  ansible.windows.win_path:
    name: PATH
    elements:
      - C:\MyCustomPath
    prepend: yes
```

### DIFF
--- a/plugins/modules/win_path.ps1
+++ b/plugins/modules/win_path.ps1
@@ -81,18 +81,18 @@ Function Add-Element ($existing_elements, $elements_to_add, $insert_at) {
                 # if new element then add and reorder in place
                 If ($idx -eq -1) {
                     $existing_elements.Insert($insert_anchor_idx, $el) | Out-Null
-                    $insert_anchor_idx++;
+                    $insert_anchor_idx++
                     $changed = $true
                 }
                 ElseIf ($idx -ne $insert_anchor_idx) {
                     # element exists, but a relative reorder is needed
                     $existing_elements.RemoveAt($idx) | Out-Null
                     $existing_elements.Insert($insert_anchor_idx, $el) | Out-Null
-                    $insert_anchor_idx++;
+                    $insert_anchor_idx++
                     $changed = $true
                 }
                 Else {
-                    $insert_anchor_idx++;
+                    $insert_anchor_idx++
                 }
             }
         }


yangskyboxlabs:
All specified elements being moved to the beginning when `prepend` feels like a surprising behaviour.  I see that this is the implementation suggested in the issue, but it does not align with the existing behaviour of not moving elements when they are already in the correct relative position to each other.

What would make more sense to me, personally, is being able to specify whether _new_ elements are appended or prepended.  They would then act as anchors used to reorder existing elements.

yangskyboxlabs:
Also, instead of a boolean parameter, it should be an enum.  Maybe `insert_at: 'end' | 'start'`.

PR #856, opened 2025-12-29
Improving Idempotency aligning options - win_file

### LINKED ISSUE #798: Enable idempotent file creation with win_file
The file module already allows for idempotent file creation so it's possible to use touch with it to only create a new file if not present already. Subsequent runs won't report "changed" if the file exists and its times are not updated. This would be helpful also with win_file.

bishalprasad321:
##### SUMMARY

This PR adds the parameters `access_time`, `access_time_format`, `modification_time`, and `modification_time_format` to the `win_file` module.  
The change improves idempotency when managing file and directory timestamps on Windows and brings `win_file` into functional parity with `ansible.builtin.file`.

Fixes #798 
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`win_file`

##### ADDITIONAL INFORMATION

With this change, users can explicitly manage access and modification times for files and directories on Windows targets.  
This enables predictable, idempotent behavior when touching or modifying filesystem objects, aligning the Windows implementation with the behavior already available on Unix-like systems via `ansible.builtin.file`.

### DIFF
--- a/plugins/modules/win_file.py
+++ b/plugins/modules/win_file.py
@@ -47,7 +47,7 @@ options:
     description:
       - The format to use when parsing C(modification_time).
       - Defaults to C(yyyy-MM-dd HH:mm:ss).
-      - See L(.NET DateTime format strings,https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings) for more information.
+      - See L(.NET DateTime format strings,https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings).
     type: str
     default: yyyy-MM-dd HH:mm:ss
     version_added: 3.4.0
@@ -65,7 +65,7 @@ options:
     description:
       - The format to use when parsing C(access_time).
       - Defaults to C(yyyy-MM-dd HH:mm:ss).
-      - See L(.NET DateTime format strings,https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings) for more information.
+      - See L(.NET DateTime format strings,https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings).
     type: str
     default: yyyy-MM-dd HH:mm:ss
     version_added: 3.4.0


bishalprasad321:
This PR is ready to be reviewed!

plugins/modules/win_file.ps1
```diff
@@ -26,6 +30,55 @@ $result = @{
...
+        $access_time = "preserve"
+    }
+}
+
+# validate correct values of mtime and atime
+$validSpecial = @("now", "preserve")
+
+if ($access_time -notin $validSpecial) {
+    try {
+        [datetime]::ParseExact($access_time, $access_time_format, $null) | Out-Null
```
jborean93: Up to you whether you want to change it or not but you could look at changing the logic so you only parse it once like

```powershell
$updateTimestamp = @{
    ...
    AccessTime = $null  # null means preserve/don't both checking
}
if ($access_time -eq "now") {
    $updateTimestamp.AccessTime = Get-Date
}
elseif ($access_time -ne "preserve") {
    try {
        $updateTimestamp.AccessTime = [datetime]::ParseExact($access_time, $access_time_format, $null)
    }
    catch {
        Fail-Json $result "Invalid access_time '$($access_time)'"
    }
}
```

You would have to update `Update-Timestamp` to make `AccessTime` and `ModificationTime` `[Nullable[DateTime]]$AccessTime` so it allows null.

The benefits of this approach are

+ You don't need to re-parse the DateTime string again
+ You don't need to provide the format value to the function
+ The checks don't need to re-check the `*_time` values and would just do something like

```powershell
# preserve will be $null here so the first check means the comparison
# is only done if set to now or a specific value.
if ($AccessTime -and $AccessTime -ne $file.LastAccessTime) {
    if (-not $CheckMode) {
        $file.LastAccessTime = $AccessTime
    }
    $changed = $true
}

File: plugins/modules/win_file.py [hunk 2]
```diff
@@ -32,6 +32,30 @@
         modification times (similar to the way C(touch) works from the command line).
     type: str
     choices: [ absent, directory, file, touch ]
+  modification_time:
+    description:
+      - The desired modification time for the file or directory.
+      - Should be V(preserve) when no modification is required, C(yyyy-MM-dd HH:mm:ss) when using default time format, or V(now).
```
jborean93: I would try and adjust the middle point to something more like  "a DateTime string in the format specified by `O(modification_time_format)`". Same for `access_time`.

plugins/modules/win_file.py
```diff
@@ -32,6 +32,30 @@
...
+      - Default is Null meaning that V(preserve) is the default for O(state=[file,directory]) and V(now) is default for O(state=touch).
```
jborean93: `Null` is a probably a bit too technical, maybe

```suggestion
      - When unset, the default is V(preserve) when O(state=[file,directory]) and V(now) when O(state=touch).
```

plugins/modules/win_file.py
```diff
@@ -62,4 +86,25 @@
...
+    path: C:\Temp\foo.conf
+    state: touch
+    modification_time: now
+    access_time: now
+
+- name: Set specific modification and access times for a file
+  ansible.windows.win_file:
+    path: C:\Temp\foo.conf
+    state: touch
+    modification_time: 20251229T123456
```
jborean93: Isn't the format meant to be `yyyy-MM-dd HH:mm:ss`, this is `yyyyMMddTHHmmss`.

plugins/modules/win_file.py
```diff
@@ -62,4 +86,25 @@
...
+    access_time: 20251229T123456
+
+- name: Create a directory and set timestamps
```
jborean93: Maybe expand and say set the timestamps to now.

plugins/modules/win_file.ps1
jborean93: A separate point to parsing once, do you know this is affected by timezones. Is the DT parsed as a UTC value, the local value, something else? Can the DT be parsed in a specific timezone by changing the format to something that would accept it. We should document this behaviour to avoid surprising anybody using this option.

tests/integration/targets/win_file/tasks/main.yml
```diff
@@ -16,7 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
-- set_fact: output_file={{ remote_tmp_dir }}\\foo.txt
+- set_fact: output_file={{ remote_tmp_dir }}/foo.txt
```
jborean93: Did you change this for a reason, did the code fail with what was there before?

plugins/modules/win_file.py
```diff
@@ -32,6 +32,30 @@
         modification times (similar to the way C(touch) works from the command line).
     type: str
     choices: [ absent, directory, file, touch ]
+  modification_time:
+    description:
+      - The desired modification time for the file or directory.
+      - Should be V(preserve) when no modification is required, C(yyyy-MM-dd HH:mm:ss) when using default time format, or V(now).
+      - Default is Null meaning that V(preserve) is the default for O(state=[file,directory]) and V(now) is default for O(state=touch).
+    type: str
```
jborean93, line 44: All these options will have to have a `version_added: 3.4.0` entry to show when the new options were added.

plugins/modules/win_file.ps1
bishalprasad321: Thanks for sharing the efficient approach, appreciate it. I've updated the requested items in my latest commit, have a review!

plugins/modules/win_file.py
bishalprasad321: This is done!

plugins/modules/win_file.py
bishalprasad321: Thankyou, updated it!

plugins/modules/win_file.py
bishalprasad321: Yeah, this was missed. Thanks - I've updated it!

plugins/modules/win_file.ps1
bishalprasad321: I've tried to include the parsing logic to be explicit, and currently the module will expect the timezone to be parsed in local time on the windows hosts. 

Tried to document about the same, have a review!

plugins/modules/win_file.py
bishalprasad321: Done!

tests/integration/targets/win_file/tasks/main.yml
bishalprasad321: I guess, it was non-related change which I introduced by mistake while writing my tests. It's reverted now,

plugins/modules/win_file.py
```diff
@@ -32,6 +35,38 @@
         modification times (similar to the way C(touch) works from the command line).
     type: str
     choices: [ absent, directory, file, touch ]
+  modification_time:
+    description:
+      - The desired modification time for the file or directory.
+      - A DateTime string in the format specified by O(modification_time_format).
+      - The timestamp is interpreted as local time on the target system.
+      - Timezone offsets are not supported.
```
jborean93: It looks like they are if the format includes the TZ specifier. For example on my host it is UTC+10 and this is what happens

```powershell
$d = [DateTime]::ParseExact('2026-01-14 11:12:01+02', 'yyyy-MM-dd HH:mm:sszz', $null)
$d.Kind
# Local
$d
# Wednesday, January 14, 2026 7:12:01 PM
```

It added the 8 hours and changed the kind to `Local` so that's correct. The docs for [DateTime.ParseExact](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.parseexact?view=net-10.0#system-datetime-parseexact(system-string-system-string-system-iformatprovider)) also documents this behaviour

> If s does represent the time in a particular time zone and format allows time zone information to be present (for example, if format is equal to the "o", "r", or "u" standard format specifiers, or if it contains the "z", "zz", or "zzz" custom format specifiers), the [Kind](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.kind?view=net-10.0) property of the returned [DateTime](https://learn.microsoft.com/en-us/dotnet/api/system.datetime?view=net-10.0) value is [DateTimeKind.Local](https://learn.microsoft.com/en-us/dotnet/api/system.datetimekind?view=net-10.0#system-datetimekind-local).

If no TZ was specified in the string or our format value didn't support a TZ then it'll return the DT as is and set the kind to be `Unspecified`. Luckily your `SpecifyKind` call afterwards just makes that a `Local` value.

plugins/modules/win_file.py
```diff
@@ -62,4 +98,25 @@
...
+    state: touch
+    modification_time: now
+    access_time: now
+
+- name: Set specific modification and access times for a file
+  ansible.windows.win_file:
+    path: C:\Temp\foo.conf
+    state: touch
+    modification_time: "2025-12-29 12:34:56"
+    access_time: "2025-12-29 12:34:56"
```
jborean93, line 116: Would be good to set an example with a custom time format that specified the timezone as UTC like

```yaml
- name: Set specific modification as UTC datetime
  ansible.windows.win_file:
    path: C:\Temp\foo.conf
    state: touch
    modification_time: "2025-12-29 12:34:56 +0"
    modification_time_format: yyyy-MM-dd HH:mm:ss z
```

We can see this works with

```powershell
$d = [DateTime]::ParseExact('2025-12-29 12:34:56 +0', 'yyyy-MM-dd HH:mm:ss z', $null)
$d.Kind  # Local
$d  # DateTime as the Local equivalent of that UTC time
```

plugins/modules/win_file.py
```diff
@@ -32,6 +35,38 @@
...
+      - A DateTime string in the format specified by O(modification_time_format).
+      - The timestamp is interpreted as local time on the target system.
+      - Timezone offsets are not supported.
+      - When unset, the default is V(preserve) when O(state=[file, directory]) and V(now) when O(state=touch).
+    type: str
+    version_added: 3.4.0
+  modification_time_format:
+    description:
+      - The format to use when parsing C(modification_time).
+      - Defaults to C(yyyy-MM-dd HH:mm:ss).
```
jborean93, line 49: Might be nice to add something like this to both this option and the acces time one

```yaml
- See L(.NET DateTime format strings,https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings)
```

plugins/modules/win_file.py
```diff
@@ -11,6 +11,9 @@
 description:
      - Creates (empty) files, updates file modification stamps of existing files,
        and can create or remove directories.
+     - Timestamp values are interpreted as local time on the target Windows system.
+     - Internally, Windows stores file timestamps in UTC, but this module operates exclusively using
```
jborean93: I think this is mostly an implementation detail and can be removed.

plugins/modules/win_file.py
```diff
@@ -11,6 +11,9 @@
 description:
      - Creates (empty) files, updates file modification stamps of existing files,
        and can create or remove directories.
+     - Timestamp values are interpreted as local time on the target Windows system.
```
jborean93: Probably best to clarify if not tz is specified.

```suggestion
     - Timestamp values are interpreted as local time on the target Windows system if no time zone offset is specified.
```

plugins/modules/win_file.ps1
```diff
@@ -89,18 +148,57 @@ function Remove-Directory($directory, $checkmode) {
     Remove-Item -LiteralPath $directory.FullName -Force -Recurse -WhatIf:$checkmode
 }
 
+# TODO: capability to support parsing in different timezones for atime/mtime
```
jborean93:
```suggestion
```

This can be done, see my other comments.

plugins/modules/win_file.py
bishalprasad321: Added an extra example for this, thankyou!

plugins/modules/win_file.py
bishalprasad321: Thanks for clarifying this, helped me to learn this behaviour!

bishalprasad321:
> Looks great, just some minor documentation suggestions around the TZ handling. Thanks for checking all this, this should be the last round of reviews.

@jborean93 Done :)

jborean93 (approved):
Looks great thanks

PR #857, opened 2026-01-12
Fixing reboot exit code

MOBergeron:
##### SUMMARY

When the shutdown binary is not found, `LASTEXITCODE` is `null` and crashes Python when attempting to convert `rc` to an `int`. With the added PowerShell condition, it ensures that the `rc` is at least set to something. Doing so, the `stderr` displayed in Ansible is the correct one instead of a Python error.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Plugin utils: `_reboot.py`

##### ADDITIONAL INFORMATION

It crashed using the Incus connection because of an [issue in Incus](https://github.com/lxc/incus/pull/2816) which could not execute an `.exe` file. Because of that, I was able to catch this bug and fix it. To reproduce, you would need to not have the `PATHEXT` environment variable or change the value to remove `.EXE`.

### DIFF
--- a/changelogs/fragments/857-win_reboot_exit_handling.yml
+++ b/changelogs/fragments/857-win_reboot_exit_handling.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - win_reboot - fix unhandled error when `.exe` not present in `PATHEXT` environment variable
+  - win_reboot - fix unhandled error when ``.exe`` not present in ``PATHEXT`` environment variable


jborean93:
Thanks for picking this up and coming up with the fix. Would you be able to add a changelog fragment for this bugfix. Basically a yaml file in https://github.com/ansible-collections/ansible.windows/tree/main/changelogs/fragments that's something like this (or whatever you feel is appropriate).

```yaml
bugfixes:
  - win_reboot - fix unhandled error when `.exe` not present in `PATHEXT` environment variable
```

changelogs/fragments/857-win_reboot_exit_handling.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - win_reboot - fix unhandled error when `.exe` not present in `PATHEXT` environment variable
```
jborean93: Sorry I forgot this is RST formatted so just need to double up on the backticks for highlighting.

```suggestion
  - win_reboot - fix unhandled error when ``.exe`` not present in ``PATHEXT`` environment variable
```

PR #860, opened 2026-02-02
dsc3 - Add support for local config file paths

jborean93:
##### SUMMARY
Add support for using a config file from a local/controller path. The default has changed to default to a local path to align with how other modules work and the `remote_config_file` can be used to control whether the local or remote path is used.

No changelog fragment is needed because the `dsc3` module has not been introduced in any version just yet.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/plugins/action/dsc3.py
+++ b/plugins/action/dsc3.py
@@ -15,7 +15,7 @@ class ActionModule(ActionBase):
     def run(
         self,
         tmp: str | None = None,
-        task_vars: dict[str, object] | None =None,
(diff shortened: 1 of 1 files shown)


jborean93:
cc @yangskyboxlabs curious what your thoughts are on this. I thought it best to add this as it lines up with how other modules that support a local or remote path. This way the `dsc3` action plugin can handle the lifetime of the remote file and deal with the copying/deleting for you.

PR #878, opened 2026-03-25
setup: Update Interface Fact Gathering

bnassif:
##### SUMMARY
This pull request changes the underlying fact gathering on interfaces to align more closely to Unix interface facts in [ansible.builtin.setup](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/setup.py).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`setup`

##### ADDITIONAL INFORMATION
The main change in this PR is allowing fact discovery against disabled interfaces, or interfaces that are part of a bond/vSwitch by changing interface discovery from `[System.Net.NetworkInformation.NetworkInterface]::GetAllNetworkInterfaces()` to `Get-NetAdapter -IncludeHidden`.

_**NOTE**: Only interfaces with MAC addresses are reported to avoid bloating with virtual interfaces_

Additionally, more facts are returned for interfaces along with those already returned:
- Added `active` attribute to interface facts to determine whether an interface is enabled
- Added `device` attribute derived from the CIM_NetworkAdapater `ifName` attribute
- Added `device_id` attribute derived from the CIM_NetworkAdapater `DeviceID` attribute
- Added `promisc` attribute derived from the CIM_NetworkAdapater `ifName` attribute
- Added `type` attribute derived from the CIM_NetworkAdapater `InterfaceType` attribute

###### Optional Addition
Root-level facts for interfaces have also been added in this PR to move closer to Unix facts.  
Full alignment will require a major change of the `ansible_facts.interfaces` variable to just be a list of interface names (`CIM_NetworkAdapter.ifName`).

This behavior can be removed from the PR if it's not desirable (L#829 - L#832).

##### OUTPUT COMPARISON
The comparison below executes the whole `Code` block in the `interface` subset, and then outputs it via:

```powershell
foreach ($interface in $formattedNetCfg) {
    $interface | Format-Table
}
```

###### BEFORE
```
Name                           Value                                                                                                                                                                                  
----                           -----                                                                                                                                                                                  
dns_domain                                                                                                                                                                                                            
connection_name                Ethernet 2                                                                                                                                                                             
default_gateway                10.172.123.1                                                                                                                                                                           
mtu                            1500                                                                                                                                                                                   
interface_name                 Red Hat VirtIO Ethernet Adapter #3                                                                                                                                                     
ipv6                           @{address=fe80::220a:171e:f336:c42d%5; prefix=64}                                                                                                                                      
macaddress                     00:00:0A:1C:58:4A                                                                                                                                                                      
speed                          10000                                                                                                                                                                                  
interface_index                5                                                                                                                                                                                      
ipv4                           @{address=10.172.123.13; prefix=25}
```

###### AFTER
```
Name                           Value
----                           -----
ipv6                           @{address=fe80::220a:171e:f336:c42d%5; prefix=64}
device                         ethernet_32774
device_id                      {5247C5F0-9077-4BB6-95C1-478AB59A7045}
connection_name                Ethernet 2
speed                          10000
mtu                            1500
default_gateway                10.172.123.1
interface_name                 Red Hat VirtIO Ethernet Adapter #3
type                           Ethernet
macaddress                     00:00:0A:1C:58:4A
ipv4                           @{address=10.172.123.13; prefix=25}
promisc                        False
dns_domain
interface_index                5
active                         True



Name                           Value
----                           -----
ipv6
device                         ethernet_32773
device_id                      {2E25FE1B-CCD0-4DF6-A0E0-6B1943D396AB}
connection_name                Ethernet
speed                          0
mtu                            0
default_gateway
interface_name                 Red Hat VirtIO Ethernet Adapter #2
type                           Ethernet
macaddress                     00:00:45:1C:58:4A
ipv4
promisc                        False
dns_domain
interface_index                3
active                         False
```

##### TIMING COMPARISONS
These changes increase execution time slightly, but offer much needed improved visibility into available interfaces and their configurations.

The comparisons below execute the whole `Code` block in the `interface` subset within `Measure-Command`:

###### BEFORE
```
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 36
Ticks             : 363907
TotalDays         : 4.21188657407407E-07
TotalHours        : 1.01085277777778E-05
TotalMinutes      : 0.000606511666666667
TotalSeconds      : 0.0363907
TotalMilliseconds : 36.3907
```

###### AFTER
```
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 153
Ticks             : 1536000
TotalDays         : 1.77777777777778E-06
TotalHours        : 4.26666666666667E-05
TotalMinutes      : 0.00256
TotalSeconds      : 0.1536
TotalMilliseconds : 153.6
```

### DIFF
--- a/changelogs/fragments/updated_interface_facts.yml
+++ b/changelogs/fragments/updated_interface_facts.yml
@@ -1,6 +1,6 @@
 ---
 major_changes:
-  - ansible.windows.setup - Changed interface discovery command from `[System.Net.NetworkInformation.NetworkInterface]::GetAllNetworkInterfaces()` to `Get-NetAdapter -IncludeHidden`
+  - ansible.windows.setup - Changed interface discovery command to report on disabled interfaces
 minor_changes:
   - ansible.windows.setup - Added `active` attribute to interface facts to determine whether an interface is enabled
   - ansible.windows.setup - Added `device` attribute derived from the CIM_NetworkAdapater `ifName` attribute
--- a/changelogs/fragments/updated_interface_facts.yml
+++ b/changelogs/fragments/updated_interface_facts.yml
@@ -1,0 +1,12 @@
+---
+major_changes:
+  - ansible.windows.setup - Changed interface discovery command from `[System.Net.NetworkInformation.NetworkInterface]::GetAllNetworkInterfaces()` to `Get-NetAdapter -IncludeHidden`
+minor_changes:
+  - ansible.windows.setup - Added `active` attribute to interface facts to determine whether an interface is enabled
+  - ansible.windows.setup - Added `device` attribute derived from the CIM_NetworkAdapater `ifName` attribute
+  - ansible.windows.setup - Added `device_id` attribute derived from the CIM_NetworkAdapater `DeviceID` attribute
+  - ansible.windows.setup - Added `promisc` attribute derived from the CIM_NetworkAdapater `ifName` attribute
+  - ansible.windows.setup - Added `type` attribute derived from the CIM_NetworkAdapater `InterfaceType` attribute
+  - ansible.windows.setup - Added root level facts for each interface, similar to how Unix interface facts are stored (i.e. `ansible_facts.{device}`)
+bufixes:
+  - ansible.windows.setup - Fixed behavior that would not report on interfaces that are disabled or part of a bond/vSwitch
--- a/changelogs/fragments/updated_interface_facts.yml
+++ b/changelogs/fragments/updated_interface_facts.yml
@@ -7,6 +7,5 @@ minor_changes:
   - ansible.windows.setup - Added `device_id` attribute derived from the CIM_NetworkAdapater `DeviceID` attribute
   - ansible.windows.setup - Added `promisc` attribute derived from the CIM_NetworkAdapater `ifName` attribute
   - ansible.windows.setup - Added `type` attribute derived from the CIM_NetworkAdapater `InterfaceType` attribute
-  - ansible.windows.setup - Added root level facts for each interface, similar to how Unix interface facts are stored (i.e. `ansible_facts.{device}`)
-bufixes:
+bugfixes:
   - ansible.windows.setup - Fixed behavior that would not report on interfaces that are disabled or part of a bond/vSwitch
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -753,83 +753,77 @@ $factMeta = @(
             $allRoutes = @(Get-NetRoute -ErrorAction SilentlyContinue)
 
             $formattedNetCfg = @(foreach ($interface in $interfaces) {
-                if ($interface.MacAddress -eq "") {
-                    continue
-                }
+                    if ($interface.MacAddress -eq "") {
+                        continue
+                    }
 
-                $interfaceType = [System.Net.NetworkInformation.NetworkInterfaceType]($interface.InterfaceType)
+                    $interfaceType = [System.Net.NetworkInformation.NetworkInterfaceType]($interface.InterfaceType)
 
-                $addresses = @($allAddresses | Where-Object InterfaceIndex -eq $interface.InterfaceIndex)
+                    $addresses = @($allAddresses | Where-Object InterfaceIndex -eq $interface.InterfaceIndex)
 
-                $ipv4_addresses = $addresses | Where-Object AddressFamily -eq "IPv4" |
-                        Select-Object @{ N = 'address'; E = { $_.IPAddress.ToString() } },
-                                    @{ N = 'prefix';  E = { $_.PrefixLength.ToString() } }
+                    $ipv4_addresses = $addresses | Where-Object AddressFamily -eq "IPv4" |
+                        Select-Object @{ N = 'address'; E = { $_.IPAddress.ToString() } }, @{ N = 'prefix'; E = { $_.PrefixLength.ToString() } }
 
-                $ipv6_addresses = $addresses | Where-Object AddressFamily -eq "IPv6" |
-                        Select-Object @{ N = 'address'; E = { $_.IPAddress.ToString() } },
-                                    @{ N = 'prefix';  E = { $_.PrefixLength.ToString() } }
+                    $ipv6_addresses = $addresses | Where-Object AddressFamily -eq "IPv6" |
+                        Select-Object @{ N = 'address'; E = { $_.IPAddress.ToString() } }, @{ N = 'prefix'; E = { $_.PrefixLength.ToString() } }
 
-                $defaultGateway = $null
+                    $defaultGateway = $null
 
-                $ipv4Default = $allRoutes |
-                    Where-Object {
-                        $_.InterfaceIndex -eq $interface.InterfaceIndex -and
-                        $_.DestinationPrefix -eq '0.0.0.0/0' -and
-                        $_.NextHop -ne '0.0.0.0'
-                    } |
-                    Sort-Object RouteMetric, ifMetric |
-                    Select-Object -First 1
-
-                if ($ipv4Default) {
-                    $defaultGateway = $ipv4Default.NextHop.ToString()
-                }
-                else {
-                    $ipv6Default = $allRoutes |
+                    $ipv4Default = $allRoutes |
                         Where-Object {
                             $_.InterfaceIndex -eq $interface.InterfaceIndex -and
-                            $_.DestinationPrefix -eq '::/0' -and
-                            $_.NextHop -ne '::'
+                            $_.DestinationPrefix -eq '0.0.0.0/0' -and
+                            $_.NextHop -ne '0.0.0.0'
                         } |
                         Sort-Object RouteMetric, ifMetric |
                         Select-Object -First 1
 
-                    if ($ipv6Default) {
-                        $defaultGateway = $ipv6Default.NextHop.ToString()
+                    if ($ipv4Default) {
+                        $defaultGateway = $ipv4Default.NextHop.ToString()
                     }
-                }
+                    else {
+                        $ipv6Default = $allRoutes |
+                            Where-Object {
+                                $_.InterfaceIndex -eq $interface.InterfaceIndex -and
+                                $_.DestinationPrefix -eq '::/0' -and
+                                $_.NextHop -ne '::'
+                            } |
+                            Sort-Object RouteMetric, ifMetric |
+                            Select-Object -First 1
 
-                $dnsClient = $allDnsClients | Where-Object InterfaceIndex -eq $interface.InterfaceIndex | Select-Object -First 1
-                $dnsDomain = if ($dnsClient -and $dnsClient.ConnectionSpecificSuffix) {
-                    $dnsClient.ConnectionSpecificSuffix
-                }
-                else {
-                    $null
-                }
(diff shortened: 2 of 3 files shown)


changelogs/fragments/updated_interface_facts.yml
```diff
@@ -0,0 +1,11 @@
+---
+major_changes:
+  - ansible.windows.setup - Changed interface discovery command to report on disabled interfaces
```
jborean93, line 3: Typically we just lump most changes in `minor_changes` and leave major for some really big things collection wide but as fact gathering is somewhat central to what people might use on Windows it's fine to leave this here.

jborean93 (approved):
Thanks for working on this, the changes are nice and clean and works just fine.

I just need to do some manually testing to see what happens when running facts with a non-admin user. I know in the past we've avoided using WMI because it may not allow non-admin users by default over a network logon but I'm unsure if that also applies to the StandardCimv2 namespace that cmdlets like `Get-NetAdapter` uses.

changelogs/fragments/updated_interface_facts.yml
Akasurde: Could you please add integration tests for this change? Thanks

jborean93: We spoke about it internally a bit and don't think you need to test that actual value, just more the structure that is returned, e.g. `ansible_facts.foo is defined` and the like.

bnassif: Thanks for the feedback. I added integration tests for the new interface facts in commit [f3d035d](https://github.com/ansible-collections/ansible.windows/pull/878/commits/f3d035d1bf293f364fbebd9e248eb7dfd765e2e5), as well as some old ones that didn't have tests defined.

Please let me know if anything else is needed!

jborean93 (changes requested):
Unfortunately I've just had a check on a test VM and it does seem like this breaks the non-admin scenario. Anything that uses WMI/CIM fails for non-admins because they are not members of the `Administrators` or `INTERACTIVE` group that these namespaces require by default. Even worse is that it slows down fact gathering altogether as it isn't an immediate failure and takes 5-10 seconds before it fails.

While the use case is niche it is something we have supported and will need to support in the future and thus we wouldn't be able to rely on the CIM based cmdlets. We would either have to find this same information in .NET or use PInvoke/some other method to retrieve this information that works for a non-admin user.

An alternative would be to gate the extra data behind an admin check but I would be reluctant to have code that diverges like this due to the limited testing around non-admin users and the code complexity required to keep as much in sync as possible.

bnassif:
> Unfortunately I've just had a check on a test VM and it does seem like this breaks the non-admin scenario. Anything that uses WMI/CIM fails for non-admins because they are not members of the `Administrators` or `INTERACTIVE` group that these namespaces require by default. Even worse is that it slows down fact gathering altogether as it isn't an immediate failure and takes 5-10 seconds before it fails.

Oof, that is unforunate 😦

I'll look into other avenues this weekend to see if this can be pursued any further. I'd appreciate if you could email me detailed replication steps so I can try to address the issues.

> An alternative would be to gate the extra data behind an admin check but I would be reluctant to have code that diverges like this due to the limited testing around non-admin users and the code complexity required to keep as much in sync as possible.

If it was only additional fields, I'd push for that route, but returning additional interfaces depending on whether user is an admin is not desirable. If anything, maybe it could be a togglable option (i.e. `ansible_win_interface_fact_method` defaulting to 'dotnet' and allowing 'wmi' for triggering this behavior).

stevefulme1:
@bnassif — I put together a draft approach that avoids CIM cmdlets to address jborean93's non-admin concern. Posting it here in case it's useful for your next iteration.

**Approach:** Use `Win32_NetworkAdapter` (WMI class, not CIM cmdlet) as the adapter source + .NET `NetworkInterface` for IP/DNS data. Falls back to .NET-only if WMI is unavailable.

**Why this works for non-admin:**
- `Win32_NetworkAdapter` via `Get-CimInstance` uses DCOM/WMI, which is available to non-admin users (unlike the `NetAdapter` CIM provider which requires elevated access)
- If even WMI fails on locked-down systems, falls back to `.NET NetworkInterface` (existing behavior, enabled adapters only)
- `.NET NetworkInterface` API for IP configuration works without admin — same as current `main`

**What it preserves from your PR:**
- Disabled interface visibility (via WMI adapter enumeration)
- `active` fact (from `NetEnabled` property)
- `type` fact (from .NET `NetworkInterfaceType` or WMI `AdapterType`)
- `device_id` fact (from WMI `DeviceID`)

**What it drops:**
- `promisc` and `device` (these required `Get-NetAdapter` — could be added later via .NET if needed)
- `Get-NetAdapter` / `Get-NetIPAddress` / `Get-DnsClient` / `Get-NetRoute` (the CIM cmdlets that break non-admin)

**Draft implementation** (replaces the `interfaces` subset Code block in `setup.ps1`):

```powershell
@{
    Subsets = 'interfaces'
    Code = {
        # Build a lookup of all .NET interfaces by ID for IP/DNS data
        $dotnetInterfaces = @{}
        foreach ($ni in [System.Net.NetworkInformation.NetworkInterface]::GetAllNetworkInterfaces()) {
            $dotnetInterfaces[$ni.Id] = $ni
        }

        # Use WMI Win32_NetworkAdapter for disabled adapter coverage.
        # Works for non-admin users (unlike Get-NetAdapter CIM cmdlet).
        $wmiAdapters = @{}
        try {
            foreach ($adapter in (Get-CimInstance -ClassName Win32_NetworkAdapter -ErrorAction SilentlyContinue)) {
                $wmiAdapters[$adapter.GUID] = $adapter
            }
        }
        catch {
            # Fall back to .NET-only (no disabled interfaces)
        }

        if ($wmiAdapters.Count -gt 0) {
            $adapterSource = 'wmi'
            $adapterKeys = $wmiAdapters.Keys
        }
        else {
            $adapterSource = 'dotnet'
            $adapterKeys = $dotnetInterfaces.Keys
        }

        $formattedNetCfg = @(foreach ($key in $adapterKeys) {
            $wmi = $null
            $dni = $null

            if ($adapterSource -eq 'wmi') {
                $wmi = $wmiAdapters[$key]
                $dni = $dotnetInterfaces[$key]
                if ($wmi.AdapterTypeId -in @(9, 53, 71)) { continue }
                if (-not $wmi.MACAddress) { continue }
            }
            else {
                $dni = $dotnetInterfaces[$key]
                if ($dni.NetworkInterfaceType -in @('Loopback', 'Tunnel')) { continue }
            }

            # Active state
            $isActive = $false
            if ($wmi) { $isActive = [bool]$wmi.NetEnabled }
            elseif ($dni) { $isActive = $dni.OperationalStatus -eq 'Up' }

            # IP config from .NET (only available for enabled adapters)
            $ipv4_address = $null
            $ipv6_address = $null
            $defaultGateway = $null
            $dnsDomain = $null
            $mtu = $null
            $index = $null

            if ($dni) {
                $ipProps = $dni.GetIPProperties()

                $ipv4_address = $ipProps.UnicastAddresses | Where-Object {
                    $_.Address.AddressFamily -eq 2
                } | Select-Object @{ N = 'address'; E = { $_.Address.ToString() } },
                                  @{ N = 'prefix'; E = { $_.PrefixLength.ToString() } }

                $ipv6_address = $ipProps.UnicastAddresses | Where-Object {
                    $_.Address.AddressFamily -eq 23
                } | Select-Object @{ N = 'address'; E = { $_.Address.ToString() } },
                                  @{ N = 'prefix'; E = { $_.PrefixLength.ToString() } }

                if ($ipProps.GatewayAddresses) {
                    $defaultGateway = $ipProps.GatewayAddresses[0].Address.IPAddressToString
                }
                if ($ipProps.DnsSuffix) { $dnsDomain = $ipProps.DnsSuffix }

                try {
                    $ipv4Props = $ipProps.GetIPv4Properties()
                    $index = $ipv4Props.Index
                    $mtu = $ipv4Props.Mtu
                }
                catch {
                    try {
                        $ipv6Props = $ipProps.GetIPv6Properties()
                        $index = $ipv6Props.Index
                        $mtu = $ipv6Props.Mtu
                    }
                    catch { }
                }
            }

            if (-not $index -and $wmi) { $index = $wmi.InterfaceIndex }

            $mac = ""
            if ($wmi -and $wmi.MACAddress) { $mac = $wmi.MACAddress.ToUpperInvariant() }
            elseif ($dni) { $mac = ($dni.GetPhysicalAddress() -replace '(..)', '$1:').ToUpperInvariant().Trim(':') }

            $connectionName = if ($wmi) { $wmi.NetConnectionID } elseif ($dni) { $dni.Name } else { "" }
            $interfaceName = if ($wmi) { $wmi.Name } elseif ($dni) { $dni.Description } else { "" }

            $speed = $null
            if ($dni -and $dni.Speed -gt 0) { $speed = $dni.Speed / 1000 / 1000 }
            elseif ($wmi -and $wmi.Speed) { $speed = $wmi.Speed / 1000 / 1000 }

            $interfaceType = ""
            if ($dni) { $interfaceType = $dni.NetworkInterfaceType.ToString() }
            elseif ($wmi) { $interfaceType = $wmi.AdapterType }

            @{
                connection_name = $connectionName
                default_gateway = $defaultGateway
                dns_domain      = $dnsDomain
                interface_index = $index
                interface_name  = $interfaceName
                ipv4            = $ipv4_address
                ipv6            = $ipv6_address
                macaddress      = $mac
                mtu             = $mtu
                speed           = $speed
                active          = $isActive
                type            = $interfaceType
                device_id       = if ($wmi) { $wmi.DeviceID } else { $null }
            }
        })

        $ansibleFacts.ansible_interfaces = $formattedNetCfg
    }
},
```

**Trade-offs:**
- `Win32_NetworkAdapter` is deprecated by Microsoft in favor of the CIM `NetAdapter` provider, but it's stable, widely available, and works without admin — good enough for fact gathering
- Disabled interfaces will show `ipv4`/`ipv6` as `$null` since they have no IP stack — this is accurate rather than misleading
- If jborean93 has concerns about `Win32_NetworkAdapter` specifically, the `GetAdaptersAddresses` PInvoke is an alternative (I included the type definition in the full draft but didn't wire it up since WMI is simpler and sufficient)

Happy to iterate on this if the maintainers have a preferred direction. Full draft file is [here](https://gist.github.com/) if you want the complete replacement block with the PInvoke definitions included.

jborean93:
Unfortunately the CIM and WMI cmdlets still use the same backend so the same restrictions apply. The main difference between the two are the remoting mechanism used by default where CIM uses WSMan (with an option to fallback to DCOM) whereas WMI uses DCOM. As this is all happening on localhost there isn't any remoting used anyway so this point is moot.

Ultimately we can't justify using a WMI/CIM source for a fact unless it's used as a backup from another source that may have failed. It's also not possible to just try and use it and fallback to null values as the failure results in a 5 second delays for each CIM/WMI call which for a task that runs on all Windows targets isn't justifiable.

If you ever wish to replicate this failure the simplest option is to setup SSH, add the non-admin user to the `OpenSSH Users` group then connect to localhost using that account. 

image

bnassif:
Since there doesn't seem to be a great solution to this, I'm closing this PR. Thank you all for your assistance.

PR #884, opened 2026-04-11
Add win_credential_info module and document per-user storage

stevefulme1:
## Summary
- Adds new `win_credential_info` module to retrieve information about credentials stored in the Windows Credential Manager
- Updates `win_credential` module documentation to clarify that credentials are stored per-user, and that credentials created under the SYSTEM account are not visible in the Credential Manager UI
- Adds changelog fragment

Fixes #879

## Context
Users creating credentials with `become_user: System` are confused when the credential doesn't appear in the Credential Manager UI. This is expected Windows behavior — SYSTEM has its own isolated credential store — but there was no way to verify credentials programmatically and the docs didn't explain this.

### New module: `win_credential_info`
- Retrieves credential metadata (name, type, username, alias, comment, persistence, attributes) from the Windows Credential Manager
- Supports filtering by `name` (with wildcard support via `CredEnumerateW`) and `type`
- When both `name` and `type` are specified, uses `CredReadW` for exact lookup
- Does not expose secrets (passwords/PINs) — this is an info module for verification only
- Uses its own C# namespace (`Ansible.CredentialManagerInfo`) to avoid conflicts with the existing `win_credential` module's types
- Follows the same patterns as `win_certificate_info` and `win_service_info`

### Documentation update: `win_credential`
- Added note explaining per-user credential storage and SYSTEM account isolation
- Added `seealso` reference to `win_credential_info`

## Test plan
- [ ] Verify `win_credential_info` returns all credentials when no filter is specified
- [ ] Verify filtering by `name` with exact match
- [ ] Verify filtering by `name` with wildcard pattern (e.g., `*.domain.com`)
- [ ] Verify filtering by `type` alone
- [ ] Verify filtering by both `name` and `type` for exact lookup
- [ ] Verify empty result when no credentials match
- [ ] Verify credentials stored under SYSTEM are returned when run with `become_user: System`
- [ ] Verify check mode support
- [ ] Verify module fails gracefully without `become`
- [ ] Run sanity tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)


changelogs/fragments/879-win_credential_info.yml
```diff
@@ -0,0 +1,9 @@
+minor_changes:
+- win_credential_info - new module to retrieve information about credentials stored
+  in the Windows Credential Manager, including credentials stored under service
+  accounts like SYSTEM that are not visible in the Credential Manager UI
+  (https://github.com/ansible-collections/ansible.windows/issues/879).
```
jborean93: We don't need a changelog entry for new modules, it is automatically added by the changelog builder.

changelogs/fragments/879-win_credential_info.yml
```diff
@@ -0,0 +1,9 @@
...
+- win_credential - added documentation note clarifying that credentials are stored
+  per-user and that credentials created under the SYSTEM account will not appear in
+  the Credential Manager UI
+  (https://github.com/ansible-collections/ansible.windows/issues/879).
```
jborean93: We typically don't add changelog entries for doc changes.

plugins/modules/win_credential_info.ps1
```diff
@@ -0,0 +1,404 @@
...
+    }
+    supports_check_mode = $true
+}
+
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$name = $module.Params.name
+$type = $module.Params.type
+
+Add-CSharpType -AnsibleModule $module -References @'
```
jborean93: Instead of duplicating the same code from `win_credential` we should make it a common module util and just reference it in both modules. We can create C# util's without any of the PowerShell wrapping code as well. While It seems like there's a few extra methods used here we should be able to re-use all the existing PInvoke logic needed in the one central location.

plugins/modules/win_credential_info.ps1
```diff
@@ -0,0 +1,404 @@
...
+        if ($credentials.Count -gt 0) {
+            $module.Result.exists = $true
+            [array]$module.Result.credentials = $credentials | ForEach-Object {
+                ConvertTo-CredentialInfo -Credential $_
+            } | Sort-Object -Property { $_.name }
+        }
+    }
+}
+catch [InvalidOperationException] {
+    $module.FailJson("$($_.Exception.Message)")
```
jborean93: Unless we can add more context information it's better to just remove the catch entirely. Typically we use try/catch with a custom `$module.FailJson()` if the existing exception is too vague or hides context around the operation that failed.

plugins/modules/win_credential_info.py
```diff
@@ -0,0 +1,129 @@
...
+---
+module: win_credential_info
+short_description: Get information on credentials stored in the Windows Credential Manager
+description:
+- Returns information about credentials stored in the Windows Credential Manager.
+- Credentials are stored per-user. When run with C(become) and C(become_user),
+  this module queries the credential store of that user. When run as SYSTEM,
+  it queries the SYSTEM account's credential store, which is separate from any
+  interactive user's store and not visible in the Credential Manager UI.
+options:
```
jborean93, line 21: We need to document when the module was added for this collection.
```suggestion
version_added: 3.6.0
options:
```

plugins/modules/win_credential_info.py
```diff
@@ -0,0 +1,129 @@
...
+  type:
+    description:
+    - The type of credential to filter by.
+    - When specified with I(name), returns a single credential matching both
+      the name and type.
+    - When omitted, credentials of all types are returned.
+    type: str
+    choices: [ domain_certificate, domain_password, generic_certificate, generic_password ]
+notes:
+- This module requires to be run with C(become) so it can access the
```
jborean93, line 37: It might be a good idea to document this won't return the actual secret, just the credential metadata.

jborean93 (commented):
Added a few comments. We definitely will want to have some integration tests to cover the various use cases here and ensure bugs don't regress in the future.

plugins/module_utils/CredentialManager.cs
```diff
@@ -0,0 +1,482 @@
...
+using System.Text;
+
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.CredentialCreateFlags -TypeName CredentialCreateFlags
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.CredentialFlags -TypeName CredentialFlags
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.CredMarshalType -TypeName CredMarshalType
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.CredentialPersist -TypeName CredentialPersist
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.CredentialType -TypeName CredentialType
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.CredentialAttribute -TypeName CredentialAttribute
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.Credential -TypeName Credential
+//TypeAccelerator -Name Ansible.Windows.CredentialManager.Win32Exception -TypeName Win32Exception
```
jborean93, line 15: These TypeAccelerators are a feature from before modules could use `using namespace ...` (before PowerShell 5.1). I highly recommend we don't use them as they aren't thread safe and use private APIs in PowerShell to setup.

Instead in the module that uses these references we can add something like this to the top of the module where `c#.namespace.here` is the value of `namespace ...` in the C# file.

```powershell
using namespace c#.namespace.here
```

From there we can reference the types directly without the verbosity of the full namespace.

plugins/module_utils/CredentialManager.cs
```diff
@@ -0,0 +1,482 @@
...
+
+namespace Ansible.CredentialManager
```
jborean93, line 17: To avoid conflicts in the future or with anything else that could be loaded in the process we follow a convention of the namespace here being the actual collection namespace. In this case it would be

```cs
namespace ansible_collections.ansible.windows.plugins.module_utils.UtilName
```

You can see this in https://github.com/ansible-collections/ansible.windows/blob/c389189baf62e5c5ca5b7c3344d5289ffe302bdf/plugins/module_utils/_CertACLHelper.cs#L16. While this is verbose it ensures the code stays completely unique and reflective of where it is defined. With the `using namespace ...` pattern mentioned above modules don't need to know this full namespace, just the underlying types inside it.

plugins/module_utils/CredentialManager.cs
jborean93, line 1: We should call this `_CredentialManager.cs` which marks it as private. This means that people should not be using this util outside of this collection and we aren't on the hook for maintaining a public API.

plugins/modules/win_credential.ps1
```diff
@@ -4,7 +4,7 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 #AnsibleRequires -CSharpUtil Ansible.Basic
-#Requires -Module Ansible.ModuleUtils.AddType
+#AnsibleRequires -CSharpUtil ansible_collections.ansible.windows.plugins.module_utils.CredentialManager
```
jborean93, line 7: You don't have to change it but just for future reference we support relative utils so this could be

```powershell
#AnsibleRequires -CSharpUtil ..module_utils.CredentialManager
```

tests/integration/targets/win_credential_info/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
+---
+- name: ensure test dir is present
+  ansible.windows.win_file:
+    path: '{{ test_credential_dir }}'
+    state: directory
+
+- name: copy the pfx certificate
+  ansible.windows.win_copy:
+    src: '{{ role_path }}/../win_credential/files/cert.pfx'
```
jborean93, line 9: I don't think test targets can reference other targets like this. We have a way to do dependency test targets like the `setup_remote_tmp_dir` one already mentioned in `meta/main.yml`.

We could create something like `setup_credential_certificate` that does the `win_certificate_store` setup (and handler to remove on end) and set the certificate thumbprint task that. All this test (and now `win_credential`) would need to do is to add this new test target to the dependencies in `meta/main.yml`. If you are unsure I can probably push a commit to this PR that does this but don't want to do so without you knowing or asking.

tests/integration/targets/win_credential_info/tasks/tests.yml
```diff
@@ -0,0 +1,566 @@
...
+  register: info_no_secret
+  vars: *become_vars
+
+- name: assert secret exists in raw lookup but not in win_credential_info
+  assert:
+    that:
+    - cred_with_secret.exists
+    - cred_with_secret.secret != ""
+    - info_no_secret.exists
+    - "'secret' not in info_no_secret.credentials[0]"
```
jborean93, line 566: I think we can remove this, I'm not 100% sure why we are testing it as we just need to check what keys exist in the `credentials` structure returned by `win_credential_info`.

jborean93:
Looks like the recent changes have reverted back to before my first review. Can you please make sure that

+ The changelog fragmenets are removed - https://github.com/ansible-collections/ansible.windows/pull/884#discussion_r3070235047
+ We are using the shared module_util for the PInvoke code - https://github.com/ansible-collections/ansible.windows/pull/884#discussion_r3070236758
+ A shared setup_ test target that can provide this test cert instead of embedding it in both targets that require it - https://github.com/ansible-collections/ansible.windows/pull/884#discussion_r3075608746

plugins/modules/win_credential_info.ps1
```diff
@@ -0,0 +1,114 @@
...
+Function ConvertTo-CredentialOutput {
+    param($InputObject)
+
+    $info = @{
+        name = $InputObject.TargetName
+        type = $InputObject.Type.ToString()
+        username = $InputObject.UserName
+        alias = $InputObject.TargetAlias
+        comment = $InputObject.Comment
+        persistence = $InputObject.Persist.ToString()
```
jborean93, line 43: Same for this one, the `win_credential` module uses `enterprise` and `local` for these options. Since we know the enums used we could do

```suggestion
        persistence = switch ($InputObject.Persist) {
            Session { 'session' }
            LocalMachine { 'local' }
            Enterprise { 'enterprise' }
            default { 'unknown 0x{0:X8}' -f ([int]$_) }
         }
```

The documentation example will also have to change to reflect this.

plugins/modules/win_credential_info.ps1
```diff
@@ -0,0 +1,114 @@
...
+    "generic_password" = [CredentialType]::Generic
+    "generic_certificate" = [CredentialType]::GenericCertificate
+}
+
+Function ConvertTo-CredentialOutput {
+    param($InputObject)
+
+    $info = @{
+        name = $InputObject.TargetName
+        type = $InputObject.Type.ToString()
```
jborean93, line 39: I think we will want to remap these PascalCase values back to the snake_case variant that `win_credential` uses for these options

```suggestion
        type = switch ($InputObject.Type) {
            Generic { 'generic' }
            DomainPassword { 'domain_password' }
            DomainCertificate { 'domain_certificate' }
            DomainVisiblePassword { 'domain_visible_password' }
            GenericCertificate { 'generic_certificate' }
            DomainExtended { 'domain_extended' }
            default { 'unknown 0x{0:X8}' -f ([int]$_) }
        }
```

The documentation example will also have to change to reflect this.

jborean93 (changes requested):
Changes look good, only a minor thing needed for the returned `persistence` and `type` values to be `snake_case` like how the modules accept. The docs for those return values will also need to be updated to reflect that.

PR #886, opened 2026-04-16
Fix failure on processor facts for non-admin

jborean93:
##### SUMMARY
Stops the code from attempting to retrieve the WMI Win32_Processor data when running as non-admin and SMBIOS does not contain the processor core information. Previously this would cause an warning in such a situation and take extra time for WMI to fail whereas now the fact will just return `null` matching how other failures work.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/setup-admin-checks.yml
+++ b/changelogs/fragments/setup-admin-checks.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - >-
+    setup - Fix failure when attempting to retrieve ``ansible_processor_cores`` and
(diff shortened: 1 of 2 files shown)


hackworth:
@jborean93 This caused a breaking change, at least on Server 2016, `$isAdmin` is never true. Reverted the collection to 3.5.0 for now, I'll open an issue when I have a moment.

jborean93:
@hackworth thanks, there is an existing issue https://github.com/ansible-collections/ansible.windows/issues/900 and a PR to fix it https://github.com/ansible-collections/ansible.windows/pull/902. I'm going to merge it in later today and push a `.1` release so you should be able to use the latest version soon.

PR #887, opened 2026-04-19
Remove 2016 from CI tests

jborean93:
##### SUMMARY
Remove Server 2016 from CI matrix as it has been removed in Ansible devel due to its impending EOL.


jborean93:
Zuul failure is unrelated to this change and seems to be a temporary one.

PR #892, opened 2026-04-23
Allow NULL SecureStrings in win_powershell sensitive_parameters

### LINKED ISSUE #891: Allow NULL SecureStrings in win_powershell sensitive_parameters
##### SUMMARY

The `win_powershell` module allows passing `sensitive_parameters` (PSCredential and SecureString) like so:

```yaml
- ansible.windows.win_powershell:
  script: |
    <# Script Body #>
  sensitive_parameters:
    - name: ParamName
      value: '{{ variable }}'
```

however when processing the `value` key, NULLs (Python `None`) are force-cast to an empty string. This causes a SecureString object with an empty string as its protected content to be passed to the script every time. There is no ability to either pass `$null` or to omit the parameter.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

win_powershell

##### ADDITIONAL INFORMATION

Some cmdlets accept a SecureString as input from the user, but default to something when the SecureString parameter is _not_ passed in. E.g. this applies to Posh-ACMEs `New-PACertificate`: https://www.powershellgallery.com/packages/Posh-ACME/4.29.0/Content/Public%5CNew-PACertificate.ps1

Ansible should expose the possibility to omit or null a SecureString sensitive_parameter to allow fallback to a cmdlets default parameter value and behavior.

The below test task can be used to illustrate the issue:

```yaml
- ansible.windows.win_powershell:
  script: |
     Param (
       [SecureString]$Password
     )

     if ($PSBoundParameters.Keys -contains 'Password') {
       'Password parameter passed in.'
     }

     if ($Password) {
       'We got a password.'
     }

     if ($Password -is [SecureString]) {
       "Password is a SecureString of length $($Password.Length)"
     }

     if ($null -eq $Password) {
       'Password is NULL'
     }
  sensitive_parameters:
    - name: Password
      value: '{{ undefined | default(None, True) }}'
      # Or any of the following tests etc.:
      # value: '{{ undefined | default(omit, True) }}'
      # value: '{{ "" | default(None, True) }}'
```

The above scriptblock can be saved as a .ps1 and invoked without any parameter:

```
.\Test-SecureString.ps1
```

and it will print:

```
Password is NULL
```

alternatively:

```
.\Test-SecureString.ps1 -Password $Null
```

prints:

```
Password parameter passed in.
Password is NULL
```

but these behaviors are impossible to achieve within a win_powershell task.

stevefulme1:
## Summary

- Changes `sensitive_parameters.value` type from `str` to `raw` so `null` passes through instead of being coerced to an empty string
- When `value` is `null` or omitted, passes `$null` to the script parameter rather than creating an empty `SecureString`
- Removes the `required_one_of` constraint on `username`/`value` so entries with only `name` are valid (passes `$null`)

This allows PowerShell scripts to use their default parameter values when no sensitive value is intended, which was previously impossible.

Fixes #891

## Test plan

- [ ] Existing SecureString and PSCredential tests continue to pass (empty string and non-empty values)
- [ ] New test: `value: null` passes `$null` to the script parameter
- [ ] New test: omitting both `value` and `username` passes `$null`
- [ ] Verify `$PSBoundParameters` reflects the parameter as bound in both null cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelogs/fragments/891-win_powershell-null-securestring.yml
+++ b/changelogs/fragments/891-win_powershell-null-securestring.yml
@@ -1,0 +1,5 @@
+minor_changes:
+  - >-
+    win_powershell - Allow sensitive_parameters entries with neither ``value``
+    nor ``username``/``password`` specified, passing ``$null`` as the parameter
+    value instead of an empty SecureString.
--- a/plugins/modules/win_powershell.ps1
+++ b/plugins/modules/win_powershell.ps1
@@ -30,7 +30,7 @@ $spec = @{
                 name = @{ type = 'str'; required = $true }
                 username = @{ type = 'str' }
                 password = @{ type = 'str'; no_log = $true }
-                value = @{ type = 'str'; no_log = $true }
+                value = @{ type = 'raw'; no_log = $true }
             }
             mutually_exclusive = @(
                 , @('value', 'username')
--- a/plugins/modules/win_powershell.ps1
+++ b/plugins/modules/win_powershell.ps1
@@ -834,7 +834,9 @@ if ($PSVersionTable.PSVersion -lt '6.0') {
                 $null
             }
 
-            $parameters[$paramDetails.name] = $value
+            if ($null -ne $value) {
+                $parameters[$paramDetails.name] = $value
+            }
         }
     }
     if ($supportsShouldProcess) {
--- a/plugins/modules/win_powershell.ps1
+++ b/plugins/modules/win_powershell.ps1
@@ -30,7 +30,7 @@ $spec = @{
                 name = @{ type = 'str'; required = $true }
                 username = @{ type = 'str' }
                 password = @{ type = 'str'; no_log = $true }
-                value = @{ type = 'raw'; no_log = $true }
+                value = @{ type = 'str'; no_log = $true }
             }
             mutually_exclusive = @(
                 , @('value', 'username')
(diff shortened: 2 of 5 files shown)


jborean93:
Looks like you've included some extra changes that are unrelated here.

As for the changes here we shouldn't have to change the type, using `type = 'str'` still allows us to pass through null. For example I have this test module locally

```powershell
#!powershell

#AnsibleRequires -CSharpUtil Ansible.Basic

$spec = @{
    options = @{
        value = @{ type = 'str' }
    }
}

$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
$module.Result.value = $module.Params.value
$module.ExitJson()
```

I can call it without `value` set, `value:`, `value: ~`, or `value: '{{ unsetvar | default(omit) }}'` and they all result in this

```yaml
ok: [win] => 
    changed: false
    invocation:
        module_args:
            value: null
    value: null
```

Seems like we just need to remove the `required_one_of = @(, @('username', 'value'))` line and have it omit the parameter altogether if both `username` and `value` are null.

stevefulme1:
Addressed both review items:

1. **Removed unrelated commits** — Rebased onto main with a single clean commit. No more win_credential_info changes leaking in.

2. **Kept `type: str`** — Reverted the `type: raw` change per your feedback. The `value` suboption stays as `str`. The key changes are:
   - Removed `required_one_of` so neither `value` nor `username` is required
   - Changed null check to `$null -ne $paramDetails.value`
   - Returns `$null` instead of empty SecureString when nothing is specified

jborean93 (approved):
Looks like the CI failure is just GitHub being unavailable at the time. Have kicked off a new run.

PR #893, opened 2026-04-23
Add .gitleaksignore for test fixture false positive

stevefulme1:
The GenericPass456 value in win_credential_info integration tests is a hardcoded test fixture, not a real credential.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -1,0 +1,1 @@
+188788b22121f08f4d273dd27e0d86eb45439f11:tests/integration/targets/win_credential_info/tasks/tests.yml:generic-api-key:62


jborean93:
I'm pretty sure we can ignore this whole file and just add `# notsecret` to the line in question. I find that better than having to deal with yet another file in the root directory which has implications around packaging the collection and whether we should or should not skip it.

stevefulme1:
Closing — applying the fix inline with `# notsecret` in #884 instead, per jborean93's suggestion.

PR #895, opened 2026-05-15
fix: fall back to WMI when SMBIOS Manufacturer is empty

noamper:
##### SUMMARY
Adds an `IsNullOrWhiteSpace` fallback to `Win32_ComputerSystem.Manufacturer` after the SMBIOSInfo construction in `setup.ps1`'s `platform` subset. Without it, hosts whose SMBIOS table exists but has a blank Type 1 (System Information) Manufacturer string end up with `ansible_system_vendor = ""`. The existing `catch [System.ComponentModel.Win32Exception]` fallback only fires when the raw SMBIOS API call itself fails (e.g. running as non-admin), not when it succeeds with empty fields.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
setup

##### ADDITIONAL INFORMATION
The same SMBIOSInfo-succeeds-but-empty condition affects the `virtual` subset, which reads both `$bios.Manufacturer` and `$bios.Model` against `manufacturerMap` / `modelMap` to populate `ansible_virtualization_type`. Out of scope for this minimal fix, but worth a separate look — the same `IsNullOrWhiteSpace`-then-fallback pattern would apply, and the lookup tables may need new entries (e.g. `BOCHS_` Manufacturer, `BXPC*` Model glob) to actually classify the host as `kvm` or `openstack` rather than `NA`.

Below output is from an impacted vm, which was migrated/converted from VMware to OpenStack with [virt-v2v](https://libguestfs.org/virt-v2v.1.html)
```
C:\temp>powershell -noprofile
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\temp> Get-CimInstance win32_operatingsystem | Select Version,Caption

Version    Caption
-------    -------
10.0.17763 Microsoft Windows Server 2019 Datacenter


PS C:\temp> Get-CimInstance -ClassName Win32_ComputerSystemProduct # no output
PS C:\temp> Get-CimInstance -ClassName Win32_ComputerSystem | Select Manufacturer,Model

Manufacturer Model
------------ -----
BOCHS_       BXPC____
```

### DIFF
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -871,6 +871,9 @@ $factMeta = @(
         Code = {
             try {
                 $bios = New-Object -TypeName Ansible.Windows.Setup.SMBIOSInfo
+                if ([string]::IsNullOrWhiteSpace($bios.Manufacturer)) {
+                    $bios = Get-CimInstance -ClassName Win32_ComputerSystem -Property Manufacturer
+                }
             }
             catch [System.ComponentModel.Win32Exception] {
                 $bios = Get-CimInstance -ClassName Win32_ComputerSystem -Property Manufacturer
--- a/changelogs/fragments/setup-platform-empty-vendor-fallback.yml
+++ b/changelogs/fragments/setup-platform-empty-vendor-fallback.yml
@@ -1,0 +1,6 @@
+bugfixes:
+  - >-
+    setup - Fix empty ``ansible_system_vendor`` on hosts where the SMBIOS Type 1 Manufacturer
+    string is blank (e.g. some Windows guests after a V2V migration). Falls back to
+    ``Win32_ComputerSystem.Manufacturer`` when SMBIOSInfo returns an empty Manufacturer,
+    matching the existing fallback path for `Win32Exception`.
--- a/changelogs/fragments/setup-platform-empty-vendor-fallback.yml
+++ b/changelogs/fragments/setup-platform-empty-vendor-fallback.yml
@@ -3,4 +3,4 @@ bugfixes:
     setup - Fix empty ``ansible_system_vendor`` on hosts where the SMBIOS Type 1 Manufacturer
     string is blank (e.g. some Windows guests after a V2V migration). Falls back to
     ``Win32_ComputerSystem.Manufacturer`` when SMBIOSInfo returns an empty Manufacturer,
-    matching the existing fallback path for `Win32Exception`.
+    matching the existing fallback path for ``Win32Exception``.


stevefulme1:
## PR Review

**Overall: Looks good — small, well-scoped bugfix.**

### Positives

1. **Well-written PR description** — clear root cause, reproduction evidence, and explicit scoping of what's in vs. out of scope.
2. **Correct fix location** — inserting the `IsNullOrWhiteSpace` check between the `try` body and the existing `catch` mirrors the fallback semantics without changing the exception path.
3. **Changelog fragment included** — properly formatted for antsibull-changelog.
4. **Good commit message** — explains *why*, not just *what*.

### Suggestions

**1. The fallback object shape differs from `SMBIOSInfo`**

When the `IsNullOrWhiteSpace` check fires, `$bios` gets reassigned to a `Win32_ComputerSystem` CIM instance. The existing `catch` path does the same thing, so this is consistent. Only `$bios.Manufacturer` is read downstream (line ~957), and both objects have that property, so this is fine.

**2. Consider also checking `$bios.Model` (minor, could be separate PR)**

The author already flags this in the description. The `virtual` subset (line ~1135) constructs its own `SMBIOSInfo` independently and reads both `$bios.Manufacturer` and `$bios.Model` — it would hit the same empty-field problem. A companion PR addressing that subset with the same pattern would be valuable. Not a blocker for this PR.

**3. There are actually three separate `SMBIOSInfo` constructions in this file**

- Line ~598 (likely `hardware`/`bios` subset) — sets `ansible_product_name = $bios.Model`
- Line ~873 (this PR's `platform` subset)
- Line ~1135 (`virtual` subset)

Each has its own `try/catch` with the same WMI fallback. The line ~598 instance would also produce an empty `ansible_product_name` on these V2V hosts. Worth a follow-up to apply the same guard to all three sites.

**4. Nit: changelog fragment formatting**

The fragment uses single backticks around `Win32Exception` but double backticks elsewhere (`` ``Win32_ComputerSystem.Manufacturer`` ``). Should be `` ``Win32Exception`` `` for consistency with the RST convention used by this project's changelog tooling.

### Verdict

**Approve with minor nit.** The fix is correct, well-scoped, and addresses a real-world scenario. The broader empty-SMBIOS-field problem across other subsets is a valid follow-up but appropriately out of scope here.

plugins/modules/setup.ps1
```diff
@@ -871,6 +871,9 @@ $factMeta = @(
         Code = {
             try {
                 $bios = New-Object -TypeName Ansible.Windows.Setup.SMBIOSInfo
+                if ([string]::IsNullOrWhiteSpace($bios.Manufacturer)) {
```
jborean93, line 874: It's interesting that even when our SMBIOS lookup fails it's still present somewhere and the `Win32_ComputerSystem` instance has the value.

If you are willing to share this information it will be curious to see the raw SMBIOS data that Windows returns. You can run this that will output the raw data as a base64 string.

```powershell
Add-Type -TypeDefinition @"
using System;
using System.Runtime.InteropServices;

public static class FirmwareTables
{
    [DllImport("kernel32.dll", SetLastError=true)]
    public static extern int GetSystemFirmwareTable(
        int FirmwareTableProviderSignature,
        int FirmwareTableID,
        IntPtr pFirmwareTableBuffer,
        int BufferSize);
}
"@

$RSMB = 0x52534D42
$size = [FirmwareTables]::GetSystemFirmwareTable($RSMB, 0, [IntPtr]::Zero, 0)
if (-not $size) {
    throw "Failed to get SMBIOS info"
}
$buffer = [System.Runtime.InteropServices.Marshal]::AllocHGlobal($size)
$result = [FirmwareTables]::GetSystemFirmwareTable($RSMB, 0, $buffer, $size)
$raw = [byte[]]::new($size)
[System.Runtime.InteropServices.Marshal]::Copy($buffer, $raw, 0, $raw.Length)
[System.Runtime.InteropServices.Marshal]::FreeHGlobal($buffer)
[Convert]::ToBase64String($raw)
```

I definitely understand if you are not comfortable doing this.

You can also use this gist https://gist.github.com/jborean93/24fa71278a8c4c365007735fe30605ae that parses that raw data into a more structured format. I'm curious whether any of the other fields in type 1 result show the same value as `Win32_ComputerSystem` `Manufacturer` that we could use as a backup instead.

noamper: The above snippet throws because `$size` is 0. 

Similarly the gist throws `Exception calling "GetRsmbTable" with "0" argument(s): "Incorrect function"`

noamper: In hindsight "succeeds with empty fields" was probably misleading, sorry

jborean93: That's weird, if that gist or snippet was failing like that then the `setup.ps1` code should have been going to the `catch` block and wouldn't be calling `if ([string]::IsNullOrWhiteSpace($bios.Manufacturer)) {` that is being added here.

The code is essentially calling it the same way, while it doesn't throw on the first call for the size check it will throw an exception on the second because `size == 0`

https://github.com/ansible-collections/ansible.windows/blob/cfc75f0854024cf31042ab1936749abb92633c82/plugins/modules/setup.ps1#L477-L483

noamper: Thanks. Anti-malware system might be intermittently fighting us. I'm looking into that and hope to circle back before too long.

noamper: I'm no longer able to reproduce the initially reported condition. At the moment `ansible_system_vendor` is populating on the handful of V2V'd targets I have readily available. Not sure if I misread something initially or if the anti-malware system is now behaving differently.

Happy to close this PR or convert to draft or whatever you think best. Thanks for bearing with me and for all the help.

noamper: ...I do see `ansible_virtualization_type='NA'` on the V2V hosts, but that's not causing me any real trouble. Mapping the manufacturer-driven `ansible_system_vendor=bochs_` to OpenStack internally seems to meet our needs. Lacking more context I would hesitate to propose any bochs_ mapping outside our shop. So probably this can be closed, because even if some bochs_ mapping is appropriate here, I imagine that would warrant a new & cleaner PR. Thanks again for the sanity check.

jborean93: No worries, I appreciate you checking through this. I would agree that it isn't ideal to map any of these raw values as when it comes to VMs these raw SMBIOS values so I think that is better suited towards the environment in which it is being used. The benefit here is it has a better known sample of what their environment looks like and what they would like it to be mapped to.

If you come across any future issues that may warrant a closer look at this mechanism we are happy to look through it but for now I think we'll close this one. Appreciate you spending the time working on this!

PR #897, opened 2026-05-20
feat: add computername parameter and fix scope filtering for leases/reservations

### LINKED ISSUE #835: Add support for ComputerName parameter in win_dhcp_lease module
SUMMARY

The win_dhcp_lease module currently only works locally on the DHCP server. It would be useful to add support for querying a remote DHCP server using the ComputerName parameter, consistent with the native PowerShell cmdlet Get-DhcpServerv4Lease.

ISSUE TYPE

Feature Idea

COMPONENT NAME

win_dhcp_lease

ADDITIONAL INFORMATION

This feature would allow users to query DHCP leases from remote servers without having to run the playbook directly on the DHCP host.

Example usage in PowerShell:
Get-DhcpServerv4Lease -ComputerName "dhcpserver.contoso.com" -ScopeId 10.10.10.0

Proposed Ansible usage:
- name: Get DHCP leases from remote DHCP server
  ansible.windows.win_dhcp_lease:
    computer_name: "dhcpserver.contoso.com"
    scope_id: "10.10.10.0"

This would align the module more closely with the underlying Windows functionality and make it easier to manage multiple DHCP servers remotely.

benoit-ferre:
##### SUMMARY
Add support for the `computername` parameter and fix scope filtering
when searching for existing DHCP leases or reservations.

Previously, lookups were not restricted to the specified scope, which
could lead to incorrect matches when the same MAC address exists in
multiple scopes (e.g. a device with multiple reservations).

This change ensures:
- support of `computername` as an input parameter
- lookup operations are restricted to the given scope for accuracy

Fixes: https://github.com/ansible-collections/ansible.windows/issues/835

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
win_dhcp_lease

###### Problem reproduction
1. Create multiple DHCP scopes
2. Add the same MAC address in more than one scope (e.g. reservations)
3. Run the module without strict scope filtering
4. Observe that the returned lease/reservation may not belong to the expected scope

###### Behavior before
- Lookup could return an object from a different scope
- Ambiguous results when MAC exists in multiple scopes

###### Behavior after
- Lookup is strictly limited to the specified scope
- No ambiguity even if MAC address exists in multiple scopes

### DIFF
--- a/plugins/modules/win_dhcp_lease.py
+++ b/plugins/modules/win_dhcp_lease.py
@@ -87,6 +87,14 @@ options:
       - Specifies the description for reservation being created.
       - Only applicable to l(type=reservation).
     type: str
+  computer_name:
+    description:
+      - Specifies a DHCP server.
+      - You can specify an IP address or any value that resolves to an IP
+        address, such as a fully qualified domain name (FQDN), host name, or
+        NETBIOS name.
+    type: str
+
 '''
 
 EXAMPLES = r'''
--- a/plugins/modules/win_dhcp_lease.py
+++ b/plugins/modules/win_dhcp_lease.py
@@ -94,6 +94,7 @@ options:
         address, such as a fully qualified domain name (FQDN), host name, or
         NETBIOS name.
     type: str
+    version_added: 3.6.0
 
 '''
 
--- a/changelogs/fragments/win_dhcp_lease.yml
+++ b/changelogs/fragments/win_dhcp_lease.yml
@@ -1,5 +1,5 @@
 bugfixes:
-  - >-    
+  - >-
     win_dhcp_lease - Fix scope filtering when searching for existing leases or reservations:
     ensure the query is limited to the specified scope. This prevents ambiguity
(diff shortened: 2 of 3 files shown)


plugins/modules/win_dhcp_lease.py
```diff
@@ -87,6 +87,14 @@
       - Specifies the description for reservation being created.
       - Only applicable to l(type=reservation).
     type: str
+  computer_name:
+    description:
+      - Specifies a DHCP server.
+      - You can specify an IP address or any value that resolves to an IP
+        address, such as a fully qualified domain name (FQDN), host name, or
+        NETBIOS name.
+    type: str
```
jborean93, line 96:
```suggestion
    type: str
    version_added: 3.6.0
```

jborean93 (approved):
Thanks for working on this, I had to remove some trailing whitespace to solve the lint error in CI and add the `version_added` entry for this new option but the rest look good.

PR #909, opened 2026-06-08
Diff enhancement / hardening

benoit-ferre:
##### SUMMARY
This pull request includes several focused fixes and small refactors to improve robustness and clarity in the win_dhcp_lease module:

- Strictly validate and normalize MAC address input in Convert-MacAddress (support colon, dash and plain hex formats; return failure for invalid values). Invalid or unexpected MAC formats could bypass validation and cause later failures.
- Rename Compare-DhcpLease to Is-DhcpLease-Changed for clearer intent.
- Add try/catch around DHCP scope/lease/reservation queries (Get-DhcpServerv4Scope, Get-DhcpServerv4Lease, Get-DhcpServerv4Reservation) and around Set-DhcpServerv4Reservation to return meaningful failures instead of uncaught exceptions.
- Use a single normalized mac_dashed value consistently for lookups and operations to avoid mismatches.
- Improve diff-mode output by populating module.Diff.after and making module.Result.lease consistent for check mode and real runs.
- Refactor parameter variable usage (introduce lease_params where appropriate) to avoid accidental reuse of temporary containers.
- Rationale: these changes fix crashes and confusing failures when invalid MAC formats are provided, when lookups raise exceptions, or when converting and updating leases/reservations; they also make module diffs clearer for users and reviewers.

##### ISSUE TYPE
Bugfix Pull Request

##### COMPONENT NAME
win_dhcp_lease

##### ADDITIONAL INFORMATION


stevefulme1:
## Review Notes

Good improvements overall — MAC validation, try/catch hardening, and diff-mode output are all welcome. Found a few issues, two of which are bugs from the `$params` → `$lease_params` rename.

### 1. BUG: `$params` still referenced after rename to `$lease_params`

The "Lease Doesn't Exist - Create" section renames `$params` to `$lease_params`, but two references were missed:

```powershell
# Line ~425 in the PR — still uses $params
if ($dns_hostname) {
    $params.HostName = $dns_hostname    # ← should be $lease_params
}
```

```powershell
# Line ~449 in the PR — still uses $params
$params.AddressState = 'Active'         # ← should be $lease_params
```

`$params` is undefined at this point in the "lease doesn't exist" code path, so this will silently create a new `$params` variable. The `HostName` and `AddressState` values won't be passed to `Add-DhcpServerv4Lease`, causing broken behavior on new lease creation with `dns_hostname` set, and missing `AddressState`.

### 2. BUG: check_mode comparison uses `$params` hashtable against lease object properties

In several places, check_mode now calls:
```powershell
$module.Result.changed = Is-DhcpLease-Changed -Original $original_lease -Updated $params
```

But `$params` is a hashtable with keys like `ClientId`, `Description`, `Name` — it does NOT have `AddressState`, `IPAddress`, or `ScopeId`. So the comparison will compare `$null` values on the Updated side for those properties, causing false positives (always reporting `changed = $true`). This is actually no better than the original hardcoded `$true`.

To fix properly, either build a complete comparison object for check_mode, or only compare properties that are actually in `$params`.

### 3. Lease creation overwrites good data with raw params

In the lease creation path (not check_mode), this sequence runs:
```powershell
# First — correct:
$module.Result.lease = Convert-ReturnValue -Object $new_lease
$module.Diff.after = Convert-ReturnValue -Object $new_lease

# Then immediately after, in the "if ($type -eq 'lease')" block:
$module.Result.lease = $lease_params     # ← overwrites the converted value
$module.Diff.after = $lease_params       # ← overwrites with raw hashtable
```

The second write overwrites the properly formatted return value with raw `$lease_params` (which has different key names like `ClientId` vs `client_id`). This creates inconsistent return data.

### 4. Function naming: `Is-DhcpLease-Changed` violates PowerShell convention

PowerShell uses `Verb-Noun` (single hyphen). `Is-DhcpLease-Changed` has two hyphens and `Is` isn't an approved verb. Better: `Test-DhcpLeaseChanged`. The original `Compare-DhcpLease` was actually more conventional, though the added explicit `return` is a good change.

### 5. Missing `Diff.after` for `state=absent`

`Diff.before` is set when a lease exists, but `Diff.after` is never set for `state=absent`. Ansible expects `Diff.after` to show the desired end state (empty/absent). Add:
```powershell
$module.Diff.after = @{}
```
after successful removal.

### 6. Tab character and double-space in `Convert-MacAddress`

```powershell
	# colon format          ← tab indentation (rest of file uses spaces)
    elseif  ($mac -match    ← double space before (
```

### 7. Missing changelog fragment

No `changelogs/fragments/` entry. Required for merge.

### 8. PR title typo

"Diff enhancement / harnening" → should be "hardening"

---

Items 1 and 3 are the highest priority — they'll cause runtime bugs on new lease creation. Item 2 means the check_mode improvement doesn't actually improve anything yet.

benoit-ferre:
Thanks for the review, @stevefulme1.

I think the current algorithm in this module is a bit hard to read and maintain.
I also noticed another redundant `if` near point 3 in your review:

```powershell
# If lease is the desired type
if ($type -eq "lease") {
    $module.Result.changed = $true
    $module.Result.lease = $lease_params
    $module.Diff.after = $lease_params
    $module.ExitJson()
}
```

Since this code is already inside an `if ($type -eq "lease")` block, this additional check seems redundant.

What do you think about refactoring this module while preserving the current behavior?

I was thinking about something along the lines of the code below.
If you think this approach makes sense, I can work on it.

The code below is only a rough draft, so please do not spend too much time reviewing this exact version.

```powershell
Function Set-DhcpLeaseDesiredState {
    Param(
        [Parameter(Mandatory = $true)]
        [Object]$Module,

        [Parameter(Mandatory = $true)]
        [string]$Type,

        [string]$IPAddress,

        [string]$ScopeId,

        [string]$Mac,

        [int]$Duration,

        [string]$DnsHostName,

        [string]$DnsRegistrationType,

        [string]$ReservationName,

        [string]$Description,

        [Parameter(Mandatory = $true)]
        [string]$State,

        [string]$ComputerName
    )

    $context = Resolve-DhcpLeaseContext `
        -Module $Module `
        -Mac $Mac `
        -IPAddress $IPAddress `
        -ScopeId $ScopeId `
        -DnsRegistrationType $DnsRegistrationType `
        -ComputerName $ComputerName

    if ($State -eq "absent") {
        if (-not $context.LeaseExists) {
            $Module.Result.changed = $false
            $Module.Result.msg = "The lease doesn't exist."
            return
        }

        Remove-DhcpLease `
            -Module $Module `
            -CurrentLease $context.CurrentLease `
            -IsReservation $context.IsReservation `
            -CheckMode $context.CheckMode `
            -ExtraArgs $context.ExtraArgs

        return
    }

    if (-not $context.LeaseExists) {
        Assert-ScopeIdRequiredForCreate `
            -Module $Module `
            -ScopeId $ScopeId

        if ($Type -eq "lease") {
            New-DhcpLease `
                -Module $Module `
                -Mac $context.Mac `
                -IPAddress $IPAddress `
                -ScopeId $ScopeId `
                -Duration $Duration `
                -DnsHostName $DnsHostName `
                -DnsRegistrationType $context.DnsRegistrationType `
                -Description $Description `
                -CheckMode $context.CheckMode `
                -ExtraArgs $context.ExtraArgs

            return
        }

        if ($Type -eq "reservation") {
            New-DhcpReservation `
                -Module $Module `
                -Mac $context.Mac `
                -IPAddress $IPAddress `
                -ScopeId $ScopeId `
                -DnsHostName $DnsHostName `
                -ReservationName $ReservationName `
                -Description $Description `
                -CheckMode $context.CheckMode `
                -ExtraArgs $context.ExtraArgs

            return
        }
    }

    if ($context.LeaseExists -and (-not $context.IsReservation)) {
        if ($Type -eq "lease") {
            Set-DhcpLease `
                -Module $Module `
                -CurrentLease $context.CurrentLease `
                -OriginalLease $context.OriginalLease `
                -Mac $context.Mac `
                -Duration $Duration `
                -DnsHostName $DnsHostName `
                -DnsRegistrationType $context.DnsRegistrationType `
                -Description $Description `
                -CheckMode $context.CheckMode `
                -ExtraArgs $context.ExtraArgs

            return
        }

        if ($Type -eq "reservation") {
            ConvertTo-DhcpReservation `
                -Module $Module `
                -CurrentLease $context.CurrentLease `
                -OriginalLease $context.OriginalLease `
                -Mac $context.Mac `
                -ReservationName $ReservationName `
                -Description $Description `
                -HostName $DnsHostName `
                -CheckMode $context.CheckMode `
                -ExtraArgs $context.ExtraArgs

            return
        }
    }

    if ($context.LeaseExists -and $context.IsReservation) {
        if ($Type -eq "lease") {
            ConvertTo-DhcpLease `
                -Module $Module `
                -CurrentLease $context.CurrentLease `
                -OriginalLease $context.OriginalLease `
                -CheckMode $context.CheckMode `
                -ExtraArgs $context.ExtraArgs

            return
        }

        if ($Type -eq "reservation") {
            Set-DhcpReservation `
                -Module $Module `
                -CurrentLease $context.CurrentLease `
                -OriginalLease $context.OriginalLease `
                -Mac $context.Mac `
                -ReservationName $ReservationName `
                -Description $Description `
                -HostName $DnsHostName `
                -CheckMode $context.CheckMode `
                -ExtraArgs $context.ExtraArgs

            return
        }
    }
}
```

plugins/modules/win_dhcp_lease.ps1
```diff
@@ -58,29 +58,29 @@ Function Convert-MacAddress {
         [string]$mac
     )
 
-    # Evaluate Length
-    if ($mac.Length -eq 12) {
-        # Insert Dashes
-        $mac = $mac.Insert(2, "-").Insert(5, "-").Insert(8, "-").Insert(11, "-").Insert(14, "-")
-        return $mac
+	# colon format
```
Akasurde:
```suggestion
    # colon format 
```

benoit-ferre:
Finally I have rewritten the module from scratch rather than applying incremental
fixes on top of the previous version. The original code had grown difficult
to follow — approximately 300 lines of procedural logic in the main script
body, parameter-building blocks duplicated three times across the conversion
paths, inconsistent error handling, and no function decomposition — making it
hard to reason about correctness or review confidently.

The rewrite delivers all the fixes described in this PR and adds several
correctness improvements that only became apparent during the refactor:

- Consistent `Diff` and `Result.lease` population across all code paths
  including check mode
- Post-write re-reads for authoritative diff values
- Post-delete verification
- Fix for false change detection in check mode caused by MAC address case
  mismatch
- Debug messages for parameters that cannot be applied to an existing plain
  lease due to the absence of `Set-DhcpServerv4Lease` in the Windows DHCP
  module (`duration`, `description`, `dns_hostname`, `dns_regtype`)

The module is now decomposed into 17 named functions with `Get-Help`
compatible documentation blocks. No parameters, return keys or idempotent
behaviour have been changed. Existing playbooks require no modifications.

I hope this rewrite can be considered acceptable despite its scope. The intent
was never to over-engineer a simple bugfix, but the original structure made it
genuinely difficult to fix. I believe the result is easier to review, easier to extend, 
and more correct than a patched version of the original would have been. I am 
of course open to feedback and happy to adjust anything that does not align 
with the project's conventions.

benoit-ferre:
Would this pull request be acceptable in its current form, despite being a complete rewrite, or would you prefer that I revert to the original implementation?

If a rewrite is not appropriate for this PR, I can instead submit a separate PR with a new module.

That said this implementation passes all checks and preserves the behavior of the original module.

jborean93:
Unfortunately this is a pretty large change and not really something I have time to review. Even if it's more maintainable the chances of it breaking some behaviour that is not currently tested is probably too high for a module in this collection.

I would highly encourage you to split up the work into smaller PRs and fix whatever components you need.

I know that's probably not what you want to hear with the work you've put in but the risk appetite of this module is quite low due to it being supported and a critical part of the Ansible and Windows side.

benoit-ferre:
I understand your position and the concerns regarding the size of the change and the risk of unintended regressions.

In that case, I will revert to the previous PR and look at smaller, more focused changes instead.

I am also planning to work on additional win_dhcp-related functionality, such as DHCP scope management. Would contributions adding new win_dhcp modules be welcome in this repository?

Also, would it make sense to propose the code from this PR as a separate module, or is having a second module covering a similar use case something that would not be desirable for the collection?

Thank you for taking the time to review and provide feedback.

jborean93:
If it doesn't complicate things too much then adding it to this PR is probably fine but it really depends. For example if scopes are part of a lease and simple metadata then probably makes sense to have them in `win_dhcp_lease` but if they are an entirely separate DHCP thing then it would make more sense to be in a different module.

I would be very reluctant to add new modules unless they have been well tested in the real world. If you are finding you have lots of DHCP functionality that you want to add I highly recommend you build your own collection and publish it to Galaxy. If there is demand for it we can potentially look at promoting it here but I'm trying to avoid bloating this collection even more. It slows down development, puts up on the hook essentially indefinitely to maintain it, and our testing times get even longer and more expensive. I would rather have a more diverse ecosystem with multiple maintainers who set their own standards and rules when it comes to collections that suite their own purposes and not have myself be the weakest link in the change when it comes to reviews/publishing.

PR #914, opened 2026-06-19
win_updates: Add assigned_only switch for WSUS environments

tw-mloschwitz:
##### SUMMARY
In environments using WSUS, the WSUS server can pre-select updates for a given client. The client may resort to looking for assigned patches in WSUS only, avoiding to overload the WSUS server or to run into the WSUS roundtrip limit, which is set to 200. This patch adds the corresponding parameter to use when using the module.

Investigations in a local setup lead to noticing that freshly delpoyed systems, when told to install Windows updates from WSUS through Ansible, would regularly corrupt their local update database. Numerous reports all over the web confirm this behaviour (cf. https://www.codeschoepfer.de/en/installing-updates-on-windows-server-systems-using-ansible/ and others). We've seen the effect of the client showing the error message 0x80244010, indicating having exceeded the WSUS server's roundtrip limit for individual queries.

A comparison of the current update handling of win_updates with a script used locally unveiled a possibly relevant difference: Said local script would allow to limit the search focus for WSUS on updates generally pre-selected by WSUS for the specific client. I am not 100% sure that this patch will actually fix the issue of corrupted local update caches, but being able to limit the updates to download to those pre-selected by WSUS for the client is not a bad thing one way or another.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_updates

### DIFF
--- a/changelogs/fragments/win-updates-support-is-assigned-wsus.yml
+++ b/changelogs/fragments/win-updates-support-is-assigned-wsus.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - win_updates - Add the "is_assigned" parameter to only query WSUS servers for updates assigned to the specific host
--- a/changelogs/fragments/win-updates-support-is-assigned-wsus.yml
+++ b/changelogs/fragments/win-updates-support-is-assigned-wsus.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - win_updates - Add the "is_assigned" parameter to only query WSUS servers for updates assigned to the specific host
+  - win_updates - Add the ``assigned_only`` option to query WSUS servers for updates assigned to the specific host only
--- a/plugins/action/win_updates.py
+++ b/plugins/action/win_updates.py
@@ -618,6 +618,7 @@ class ActionModule(ActionBase):
 
     _VALID_ARGS = [
         'accept_list',
+        'assigned_only',
         'category_names',
         'log_path',
         'reboot',
--- a/plugins/modules/win_updates.ps1
+++ b/plugins/modules/win_updates.ps1
@@ -1751,22 +1751,47 @@ namespace Ansible.Windows.WinUpdates
 
     $query = 'IsInstalled = 0'
     if ($AssignedOnly) {
+        $useAssigned = $false
         if ($ServerSelection -eq 'managed_server') {
-            # Narrow the server-side search to updates assigned/approved to this
-            # host by the managed update server (WSUS). Mirrors a manual WUA scan
-            # of 'IsInstalled=0 AND IsAssigned=1' and avoids pulling metadata for
-            # the entire non-installed catalog.
+            $useAssigned = $true
+        }
+        elseif ($ServerSelection -eq 'default') {
+            # Under 'default' the WUA uses whatever source the host is configured for.
+            # If the host is GPO-configured for a managed server (WSUS) then IsAssigned
+            # is meaningful, so honour assigned_only. Detect via the WindowsUpdate policy
+            # keys (UseWUServer = 1 and a WUServer set).
+            try {
+                $auKey = [Microsoft.Win32.Registry]::LocalMachine.OpenSubKey('SOFTWARE\Policies\Microsoft\Windows\WindowsUpdate\AU')
+                $wuKey = [Microsoft.Win32.Registry]::LocalMachine.OpenSubKey('SOFTWARE\Policies\Microsoft\Windows\WindowsUpdate')
+                $useWUServer = if ($auKey) { $auKey.GetValue('UseWUServer') } else { $null }
+                $wuServer = if ($wuKey) { $wuKey.GetValue('WUServer') } else { $null }
+                if ($auKey) { $auKey.Dispose() }
+                if ($wuKey) { $wuKey.Dispose() }
+                if (($useWUServer -eq 1) -and (-not [string]::IsNullOrEmpty($wuServer))) {
+                    $useAssigned = $true
+                    $api.WriteLog("'assigned_only': server_selection is 'default' but the host is " +
+                        "GPO-configured for a managed server (UseWUServer=1, WUServer='$wuServer'); " +
+                        "applying 'IsAssigned = 1'.")
+                }
+            }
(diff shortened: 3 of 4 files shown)


plugins/modules/win_updates.py
```diff
@@ -45,6 +45,25 @@
...
+        - This mirrors a manual Windows Update / WUA scan and can avoid large
+          metadata transfers and errors such as C(0x80244010)
+          (WU_E_PT_EXCEEDED_MAX_SERVER_TRIPS) on hosts pointed at a WSUS server
+          that publishes a large catalog.
+        - The C(IsAssigned) criterion only has a defined meaning against a managed
+          update server. If O(server_selection) is not C(managed_server) this
+          option is ignored and a warning is written to the update log, because
+          against Windows Update / Microsoft Update it may return no updates.
+        type: bool
+        default: no
```
Akasurde, line 68:
```suggestion
        default: no
        version_added: '3.7.0'
```

plugins/modules/win_updates.ps1
```diff
@@ -1745,6 +1750,24 @@ namespace Ansible.Windows.WinUpdates
...
+            $query = "$query AND IsAssigned = 1"
+        }
+        else {
+            # IsAssigned only has a defined meaning against a managed server.
+            # Against Windows Update / Microsoft Update it can return an empty
+            # result, so ignore the option and warn instead of silently finding
+            # nothing.
+            $api.WriteLog("WARNING: 'assigned_only' is ignored because " +
+                "'server_selection' is '$ServerSelection', not 'managed_server'. " +
+                "IsAssigned only applies to a managed update server (e.g. WSUS).")
```
Akasurde:
```suggestion
                "IsAssigned only applies to a managed update server (For example, WSUS).")
```

Akasurde: helps in better translation.

plugins/action/win_updates.py
```diff
@@ -618,6 +618,7 @@ class ActionModule(ActionBase):
 
     _VALID_ARGS = [
         'accept_list',
+        'assigned_only',
```
Akasurde, line 621: Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example.

Akasurde (changes requested):
Code wise LGTM,

Akasurde:
CI Failures are unrelated to the changes and handled in PR https://github.com/ansible-collections/ansible.windows/pull/920

tw-mloschwitz:
Thank you for the review, I performed the changes as requested.

changelogs/fragments/win-updates-support-is-assigned-wsus.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - win_updates - Add the "is_assigned" parameter to only query WSUS servers for updates assigned to the specific host
```
jborean93:
```suggestion
  - win_updates - Add the ``assigned_only`` option to only query WSUS servers for updates assigned to the specific host
```

plugins/modules/win_updates.ps1
```diff
@@ -1745,6 +1750,25 @@ namespace Ansible.Windows.WinUpdates
...
+            # of 'IsInstalled=0 AND IsAssigned=1' and avoids pulling metadata for
+            # the entire non-installed catalog.
+            $query = "$query AND IsAssigned = 1"
+        }
+        else {
+            # IsAssigned only has a defined meaning against a managed server.
+            # Against Windows Update / Microsoft Update it can return an empty
+            # result, so ignore the option and warn instead of silently finding
+            # nothing.
+            $api.WriteLog("WARNING: 'assigned_only' is ignored because " +
```
jborean93: What if the `server_select: default` is set and Windows is configured to use WSUS? If the default option with a WSUS configuration on the server works then it sounds like maybe we should just add the query if set rather than only when `managed_server` is specified.

jborean93 (changes requested):

plugins/modules/win_updates.ps1
tw-mloschwitz: I have extended the code to support this.

plugins/modules/win_updates.ps1
```diff
@@ -1745,6 +1750,50 @@ namespace Ansible.Windows.WinUpdates
     $api.WriteLog("Search source set to '$($ServerSelection)' (ServerSelection = $($serverSelectionValue))")
 
     $query = 'IsInstalled = 0'
+    if ($AssignedOnly) {
+        $useAssigned = $false
+        if ($ServerSelection -eq 'managed_server') {
+            $useAssigned = $true
+        }
+        elseif ($ServerSelection -eq 'default') {
+            # Under 'default' the WUA uses whatever source the host is configured for.
```
jborean93, line 1759: This is a lot of extra work, do we really need all this? What happens if we just use `IsAssigned = 1` in the query when WU is being used? I assumed no updated are found? If so then it sounds like it's just simpler to just always apply the extra query filter if requested by the option and just make sure our documentation state it's for WSUS servers only and doing it with WU/MU will result in no updates being found.

tw-mloschwitz: Well we'll have to do it one way or another. If we want to automatically detect whether a machine is tied to a WSUS- or WSUS-like update server by default, this is roughly the code we need. If we are fine with documenting that is_assigned possibly breaks on non-wsus machines, I am fine wit that as well. What do we want to do?

jborean93: I think just document is the way to go. Adding the check at runtime, especially for a warning that is only visible in a user opt in log path isn't ideal.

stevefulme1: Is this for the standlone WSUS server or the WSUS server within MECM/SCCM? If so we have a MECM collection that may do this already.

PR #916, opened 2026-06-21
Add ansible_netbios_domain fact

dagwieers:
##### SUMMARY
This adds the ansible_netbios_domain, often called the shortname of the AD domain. When the system is not part of a domain/workgroup, this is identical to the NetBIOSName or hsotname.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
setup.ps1

### DIFF
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -70,7 +70,7 @@ using Microsoft.Win32.SafeHandles;
 using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
-using System.Security.Principal:
+using System.Security.Principal;
 using System.Text;
 
 namespace Ansible.Windows.Setup
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -70,6 +70,7 @@ using Microsoft.Win32.SafeHandles;
 using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
+using System.Security.Principal:
 using System.Text;
 
 namespace Ansible.Windows.Setup
@@ -548,6 +549,7 @@ namespace Ansible.Windows.Setup
     public class SystemInfo
     {
         public string NetBIOSName;
+        public string NetBIOSDomain;
         public UInt32 NumberOfProcessors;
         public UInt32 ProcessorArchitecture;
 
@@ -569,6 +571,7 @@ namespace Ansible.Windows.Setup
             if (!NativeMethods.GetComputerNameExW(0, buffer, ref size))
                 throw new Win32Exception("Failed to get ComputerName");
             NetBIOSName = buffer.ToString();
+            NetBIOSDomain = WindowsIdentity.GetCurrent().Name.Split('\\')[0];
         }
     }
 }
@@ -948,6 +951,7 @@ $factMeta = @(
             $ansibleFacts.ansible_fqdn = $fqdn
             $ansibleFacts.ansible_hostname = $ipProps.HostName
             $ansibleFacts.ansible_netbios_name = $systemInfo.NetBIOSName
+            $ansibleFacts.ansible_netbios_domain = SystemInfo.NetBIOSDomain
             $ansibleFacts.ansible_kernel = $osVersion.Version.ToString()
             $ansibleFacts.ansible_nodename = $fqdn
             $ansibleFacts.ansible_machine_id = $machineSid


plugins/modules/setup.ps1
```diff
@@ -70,6 +70,7 @@ using Microsoft.Win32.SafeHandles;
 using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
+using System.Security.Principal:
```
Akasurde:
```suggestion
using System.Security.Principal;
```

plugins/modules/setup.ps1
```diff
@@ -569,6 +571,7 @@ namespace Ansible.Windows.Setup
             if (!NativeMethods.GetComputerNameExW(0, buffer, ref size))
                 throw new Win32Exception("Failed to get ComputerName");
             NetBIOSName = buffer.ToString();
+            NetBIOSDomain = WindowsIdentity.GetCurrent().Name.Split('\\')[0];
```
jborean93, line 574: This is really just the user's domain portion and not really a fact about the host itself. It seems like we will need to modify the `DomainInfo` class to return this information so it's not dependent on whether you are using a local or domain user.

Luckily we already get this info and just need to surface it.

```diff
diff --git a/plugins/modules/setup.ps1 b/plugins/modules/setup.ps1
index 52205cc..4eb1d48 100644
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -296,6 +296,7 @@ namespace Ansible.Windows.Setup
     public class DomainInfo
     {
         public string Domain;
+        public string DomainNetbiosName;
         public Int32 DomainRole;
         public bool PartOfDomain;
 
@@ -313,6 +314,7 @@ namespace Ansible.Windows.Setup
                     dsBuffer.DangerousGetHandle(), typeof(NativeHelpers.DSROLE_PRIMARY_DOMAIN_INFO_BASIC));
 
                 Domain = domainInfo.DomainNameDns;
+                DomainNetbiosName = domainInfo.DomainNameFlat;
                 DomainRole = domainInfo.MachineRole;
                 PartOfDomain = !String.IsNullOrEmpty(Domain);
             }
@@ -1101,6 +1103,7 @@ $factMeta = @(
             }
 
             $ansibleFacts.ansible_windows_domain = $domainInfo.Domain
+            $ansibleFacts.ansible_windows_domain_netbios = $domainInfo.DomainNetbiosName
             $ansibleFacts.ansible_windows_domain_member = $domainInfo.PartOfDomain
             $ansibleFacts.ansible_windows_domain_role = $domainRole
         }
```

I stuck with the `ansible_windows_domain_*` prefix like the other domain facts but happy to talk further if you think it should be something else.

jborean93 (changes requested):

plugins/modules/setup.ps1
dagwieers: I used the following code to get to this information elsewhere:
```powershell
    $forest = [System.DirectoryServices.ActiveDirectory.Domain]::GetComputerDomain().Name
    $facts.netbios_domain = (Get-WmiObject -Class Win32_NTDomain -Filter "DnsForestName = '$forest'").DomainName
```

We already have **ansible_netbios_name** exposed, so it made more sense to me to name this **ansible_netbios_domain**.

dagwieers: After trying your suggested implementation, it returns null for me.

jborean93: > I used the following code to get to this information elsewhere:

We can't use WMI/CIM as when running as a non-admin user it fails but unfortunately does not fail straight away. It takes about 5-10 seconds for it finally timeout which is not a delay we want for such a common module run.

> We already have ansible_netbios_name exposed, so it made more sense to me to name this ansible_netbios_domain.

I can definitely see that but the AD domain facts are already grouped under `ansible_windows_domain_*` so it made more sense to me to try and group the domain's netbios name under that. Pedantically the `netbios` portion is a subset of the domain information rather than the domain being a subset of netbios itself.

> After trying your suggested implementation, it returns null for me.

That's interesting because I did try it out

```yaml
# ansible DC01 -m setup | grep domain

        ansible_domain: domain.test
        ansible_fqdn: DC01.domain.test
        ansible_nodename: DC01.domain.test
        ansible_os_product_type: domain_controller
        ansible_user_dir: C:\Users\vagrant-domain
        ansible_user_id: vagrant-domain
        ansible_windows_domain: domain.test
        ansible_windows_domain_member: true
        ansible_windows_domain_netbios: DOMAIN  # -- New Fact Here -- #
        ansible_windows_domain_role: Primary domain controller
```

The docs for [DSROLE_PRIMARY_DOMAIN_INFO_BASIC](https://learn.microsoft.com/en-us/windows/win32/api/dsrole/ns-dsrole-dsrole_primary_domain_info_basic) where we are getting this info do state that `DomainNameFlat` is the netbios name

> Pointer to a null-terminated Unicode string that contains the NetBIOS domain name.

PR #917, opened 2026-06-21
Ensure ansible_domain is also set when not domain-joined

dagwieers:
##### SUMMARY
Currently, we only get an ansible_domain when we are domain-joined. But this is incorrect.

The AD domain is set in ansible_windows_domain, the DNS domain is set in ansible_domain.

When domain-joined this is typically the same, but a stand-alone system should also have its domain set.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
setup.ps1


jborean93 (approved):

jborean93:
Pushed a commit to add the changelog fragment and just spaced out the if statement across multiple lines to avoid the linting error and make it easier to read. Will merge once CI is green.

PR #918, opened 2026-06-21
Ensure ansible_machine_id works when LanmanServer is stopped/disabled (hardened Windows)

dagwieers:
##### SUMMARY
On a hardened Windows OS, the LanManserver service may be stopped or disabled, this will cause an exception when trying to retrieve the machine id.

This fix will try an alternative implementation when this occurs. This was tested on a hardened server affected by this bug.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
setup.ps1

##### ADDITIONAL INFORMATION
The following exception would be thrown on a hardened system:
```powershell
    [WARNING]: Error during machine sid retrieval: Exception calling ".ctor" with "2" argument(s): "The service is not started.  "
```

### DIFF
--- a/plugins/modules/setup.ps1
+++ b/plugins/modules/setup.ps1
@@ -920,7 +920,18 @@ $factMeta = @(
                 $groupMembers.Dispose()
             }
             catch {
-                $module.Warn("Error during machine sid retrieval: $($_.Exception.Message)")
+                if ($_.Exception.ToString() -match 'Server service is not started') {
+                    $admin = Get-CimInstance Win32_UserAccount -Filter "LocalAccount=True AND SID LIKE '%-500'"
+                    if ($admin) {
+                        $machineSid = ($admin.SID -replace '-500$')
+                    }
+                    else {
+                        $module.Warn('Error retrieving machine sid from Win32_UserAccount')
+                    }
+                }
+                else {
+                    $module.Warn("Error during machine sid retrieval: $($_.Exception.Message)")
+                }
             }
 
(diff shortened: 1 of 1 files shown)


plugins/modules/setup.ps1
```diff
@@ -920,7 +920,18 @@ $factMeta = @(
                 $groupMembers.Dispose()
             }
             catch {
-                $module.Warn("Error during machine sid retrieval: $($_.Exception.Message)")
+                if ($_.Exception.ToString() -match 'Server service is not started') {
+                    $admin = Get-CimInstance Win32_UserAccount -Filter "LocalAccount=True AND SID LIKE '%-500'"
```
jborean93, line 924: I'm not sure if this is worth the lookup, knowing the machine SID is not really too useful. If we did want to go ahead with this and you have a need I would change this PR to do the following

+ See if the exception has a HResult or ErrorCode property that we can filter by
  + Filtering by the string is brittle and breaks across localisation
  + An actual error code stays constant
+ Only process with the CIM class if `$isAdmin` is true, if this runs as non-admin it will hang for a few seconds before it fails

Another option is to see if there's another API we can use instead of the one reliant on the LanmanServer service. There probably is but most likely relies on a lot more PInvoke code to reach.

Ultimately this is a lot of extra work and checks just to get this one fact, if it's not actually useful then I would be reluctant to make it even more complicated than it already is.

dagwieers: I do not disagree. If we can get the ugly error go away and just report that the machine_id cannot be retrieved due to LanmanServer not running (maybe hinting this is due to hardening) I am fine. I will redo this PR.

PR #920, opened 2026-06-22
Update test container

jborean93:
##### SUMMARY
Updates the `azure-pipelines-test-container` to the latest version so we can start using devel and newer Python versions in testing that are currently failing.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -31,7 +31,7 @@ variables:
   - name: fetchDepth
     value: 0
   - name: defaultContainer
-    value: quay.io/ansible/azure-pipelines-test-container:7.0.0
+    value: quay.io/ansible/azure-pipelines-test-container:8.0.0
 
 pool: Standard
 
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -133,6 +133,8 @@ stages:
     displayName: Ansible 2.17
     dependsOn:
       - Dependencies
+    variables:
+      defaultContainer: quay.io/ansible/azure-pipelines-test-container:7.0.0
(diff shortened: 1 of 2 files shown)


jborean93:
@briantist is there an updated template or workflow I should now be following for building and validating the docs. It seems like we were using the `push` and not `pr` workflow target in `docs-pr.yml` workflow. I've updated it here but will have to wait until it's merged into main to verify. I'm unsure if there has been any other recommended changes that should be made to ensure it is secure and continues to work.

briantist:
> @briantist is there an updated template or workflow I should now be following for building and validating the docs. It seems like we were using the `push` and not `pr` workflow target in `docs-pr.yml` workflow. I've updated it here but will have to wait until it's merged into main to verify. I'm unsure if there has been any other recommended changes that should be made to ensure it is secure and continues to work.

I think both push and pr make sense to use (one for pushes to main and one for PR docs), but I'd have to refresh my memory because it's been a long time since I've set these up. 

You can also potentially test the changes without merging by opening a new PR against this PR's branch instead of against main.

I'll tag in @felixfontein who has been doing most of the recent maintenance and improvements to these workflows.

jborean93:
Thanks I've just pushed a new change based on what community.crypto is using and after some testing in `microsoft.iis` to verify it works with the slightly different workflow we use here (we don't publish on a PR).

+ https://github.com/ansible-collections/microsoft.iis/pull/64
+ https://github.com/ansible-collections/microsoft.iis/pull/65

jborean93:
If @felixfontein recommends another method or tweak we will follow it up in another PR. For now I need to get CI in general to be green so I can start reviewing and merging some of the bugfixes and features that have been opened.

felixfontein:
The workflows look good to me!

PR #936, opened 2026-08-06
Add win_reboot_info module

mikemorency:
##### SUMMARY
Adds a new `win_reboot_info` module that returns detailed reboot status information for a Windows host, including:

- Last boot time (ISO 8601 and Unix epoch)
- Details about the most recent shutdown/restart event (Event ID 1074) — who initiated it, what process triggered it, and why
- Whether a reboot is pending, with a detailed breakdown of which sources require a reboot (Component Based Servicing, Windows Update, pending file renames, computer rename, domain join, Server Manager)

This module is designed to provide more granular reboot information than what is available through the `setup` module's `ansible_reboot_pending` fact. While there is some overlap in the pending reboot checks (both check CBS, PendingFileRename, and ServerManager), `win_reboot_info` covers additional sources (Windows Update, pending computer rename, domain join) and provides structured detail about each pending reason, as well as last reboot event context that `setup` does not offer.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/changelogs/fragments/06082026-win_reboot_info.yml
+++ b/changelogs/fragments/06082026-win_reboot_info.yml
@@ -1,4 +1,0 @@
----
-minor_changes:
-  - win_reboot_info - Add module to gather information about the last reboot and any pending
-    reboots
--- a/changelogs/fragments/06082026-win_reboot_info.yml
+++ b/changelogs/fragments/06082026-win_reboot_info.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - win_reboot_info - Add module to gather information about the last reboot and any pending
+    reboots
--- a/plugins/modules/win_reboot_info.ps1
+++ b/plugins/modules/win_reboot_info.ps1
@@ -1,0 +1,183 @@
+#!powershell
+
+# Copyright: (c) 2026, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+function Get-LastBootTime {
+    param([Ansible.Basic.AnsibleModule]$Module)
+
+    try {
+        $epochStart = New-Object -TypeName DateTime -ArgumentList @(1970, 1, 1, 0, 0, 0, [DateTimeKind]::Utc)
+        $lastBoot = (Get-CimInstance -ClassName Win32_OperatingSystem -Property LastBootUpTime).LastBootUpTime
+        $utcBoot = $lastBoot.ToUniversalTime()
+
+        return @{
+            last_boot_time = $utcBoot.ToString("yyyy-MM-ddTHH:mm:ssZ")
+            last_boot_time_epoch = (New-TimeSpan -Start $epochStart -End $utcBoot).TotalSeconds
+            utc_datetime = $utcBoot
+        }
+    }
+    catch {
+        $Module.FailJson("Failed to get last boot time from CIM: $($_.Exception.Message)")
+    }
+}
+
+function Get-LastRebootEvent {
+    param(
+        [Ansible.Basic.AnsibleModule]$Module,
+        [DateTime]$BootTime
+    )
+
+    try {
+        $rebootEvent = Get-WinEvent -FilterHashtable @{
+            LogName = 'System'
+            Id = 1074
+        } -MaxEvents 1 -ErrorAction SilentlyContinue
+
+        if ($rebootEvent) {
+            $eventTime = $rebootEvent.TimeCreated.ToUniversalTime()
+            if ($eventTime -lt $BootTime.AddMinutes(-5)) {
+                return $null
+            }
+
+            return @{
+                process = $rebootEvent.Properties[0].Value
+                reason = $rebootEvent.Properties[2].Value
+                type = $rebootEvent.Properties[4].Value
+                comment = $rebootEvent.Properties[5].Value
+                initiated_by = $rebootEvent.Properties[6].Value
+                event_time = $eventTime.ToString("yyyy-MM-ddTHH:mm:ssZ")
+            }
+        }
+    }
+    catch {
+        $Module.Warn("Failed to query the System event log for reboot events: $($_.Exception.Message)")
+    }
+
+    return $null
+}
+
+function Test-ComponentBasedServicing {
+    $cbsPath = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Component Based Servicing\RebootPending'
+    if (Test-Path -LiteralPath $cbsPath) {
+        return @{
+            source = 'component_based_servicing'
+            description = 'Component Based Servicing has a pending reboot.'
+        }
+    }
+}
+
+function Test-WindowsUpdate {
+    $wuPath = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\WindowsUpdate\Auto Update\RebootRequired'
+    if (Test-Path -LiteralPath $wuPath) {
+        return @{
+            source = 'windows_update'
+            description = 'Windows Update has installed updates that require a reboot.'
+        }
+    }
+}
+
+function Test-PendingFileRename {
+    $params = @{
+        LiteralPath = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager'
+        Name = 'PendingFileRenameOperations'
+        ErrorAction = 'SilentlyContinue'
+    }
+    $pendingRenames = (Get-ItemProperty @params)."$($params.Name)"
+    if ($pendingRenames) {
+        return @{
+            source = 'pending_file_rename'
+            description = 'There are pending file rename operations that require a reboot.'
+        }
+    }
+}
+
+function Test-PendingComputerRename {
+    $pendingParams = @{
+        LiteralPath = 'HKLM:\SYSTEM\CurrentControlSet\Control\ComputerName\ComputerName'
+        Name = 'ComputerName'
+        ErrorAction = 'SilentlyContinue'
+    }
+    $activeParams = @{
+        LiteralPath = 'HKLM:\SYSTEM\CurrentControlSet\Control\ComputerName\ActiveComputerName'
+        Name = 'ComputerName'
+        ErrorAction = 'SilentlyContinue'
+    }
+    $pendingName = (Get-ItemProperty @pendingParams)."$($pendingParams.Name)"
+    $activeName = (Get-ItemProperty @activeParams)."$($activeParams.Name)"
+    if ($pendingName -and $activeName -and $pendingName -ne $activeName) {
+        return @{
+            source = 'pending_computer_rename'
+            description = "Computer name change pending from '$activeName' to '$pendingName'."
+        }
+    }
+}
+
+function Test-DomainJoin {
+    $params = @{
+        LiteralPath = 'HKLM:\SYSTEM\CurrentControlSet\Services\Netlogon'
+        ErrorAction = 'SilentlyContinue'
+    }
+    $props = Get-ItemProperty @params
+    if ($props -and ($props.PSObject.Properties.Name -contains 'JoinDomain' -or $props.PSObject.Properties.Name -contains 'AvoidSpnSet')) {
+        return @{
+            source = 'domain_join'
+            description = 'A domain join operation is pending and requires a reboot.'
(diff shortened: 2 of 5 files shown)


changelogs/fragments/06082026-win_reboot_info.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
```
jborean93: New modules don't need a changelog fragment, they are automatically added to the changelog when the release happens.

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+        $utcBoot = $lastBoot.ToUniversalTime()
+
+        return @{
+            last_boot_time = $utcBoot.ToString("yyyy-MM-ddTHH:mm:ssZ")
+            last_boot_time_epoch = (New-TimeSpan -Start $epochStart -End $utcBoot).TotalSeconds
+            utc_datetime = $utcBoot
+        }
+    }
+    catch {
+        $Module.FailJson("Failed to get last boot time from CIM: $($_.Exception.Message)")
```
jborean93: It's best to stringify the actual `ErrorRecord` value rather than the exception message as the record could have extra details not in the exception. Also good to include the record as the second argument for `FailJson` to utilise.

```suggestion
        $Module.FailJson("Failed to get last boot time from CIM: $_", $_)
```

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+    try {
+        $rebootEvent = Get-WinEvent -FilterHashtable @{
+            LogName = 'System'
+            Id = 1074
+        } -MaxEvents 1 -ErrorAction SilentlyContinue
+
+        if ($rebootEvent) {
+            $eventTime = $rebootEvent.TimeCreated.ToUniversalTime()
+            if ($eventTime -lt $BootTime.AddMinutes(-5)) {
+                return $null
```
jborean93: It's best to just not return anything rather than `$null`. While the end result is the same here there is a slight difference in how PowerShell treats an explicit null vs implicit no return value.

```suggestion
                return
```

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+                initiated_by = $rebootEvent.Properties[6].Value
+                event_time = $eventTime.ToString("yyyy-MM-ddTHH:mm:ssZ")
+            }
+        }
+    }
+    catch {
+        $Module.Warn("Failed to query the System event log for reboot events: $($_.Exception.Message)")
+    }
+
+    return $null
```
jborean93:
```suggestion
```

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+                reason = $rebootEvent.Properties[2].Value
+                type = $rebootEvent.Properties[4].Value
+                comment = $rebootEvent.Properties[5].Value
+                initiated_by = $rebootEvent.Properties[6].Value
+                event_time = $eventTime.ToString("yyyy-MM-ddTHH:mm:ssZ")
+            }
+        }
+    }
+    catch {
+        $Module.Warn("Failed to query the System event log for reboot events: $($_.Exception.Message)")
```
jborean93:
```suggestion
        $Module.Warn("Failed to query the System event log for reboot events: $_", $_)
```

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+    If a check is null, that result is filtered out.
+    #>
+    $checks = @(
+        Test-ComponentBasedServicing
+        Test-WindowsUpdate
+        Test-PendingFileRename
+        Test-PendingComputerRename
+        Test-DomainJoin
+        Test-ServerManager
+    ) | Where-Object { $_ }
```
jborean93: The `Where-Object` shouldn't be needed AFAIK. All these values return a hashtable if it's pending and nothing if not. This nothing is called `AutomationNull` in PowerShell` and is ignored when it comes to emitting it in an expression.

For example

```powershell
function Test-AutomationNull {
    if ($args[0]) { 'value' }
    # $false emits nothing and becomes AutomationNull
}

$res = @(
    Test-AutomationNull $true
    Test-AutomationNull $false
)

ConvertTo-Json -InputObject $res -Compress
# ["value"]
```

Personally I think we could remove this entire function and just call each individual function in the `$reasons = ...` call below.

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$bootTime = Get-LastBootTime -Module $module
+$module.Result.last_reboot = @{
+    time = $bootTime.last_boot_time
+    time_epoch = $bootTime.last_boot_time_epoch
+    details = Get-LastRebootEvent -Module $module -BootTime $bootTime.utc_datetime
+}
+
+$reasons = Get-PendingRebootReason
```
jborean93: By avoiding the function we could simplify this by assigning the value directly here. No need to mess with the unary operator in the return value.
```suggestion
$reasons = @(
    Test-ComponentBasedServicing
    Test-WindowsUpdate
    Test-PendingFileRename
    Test-PendingComputerRename
    Test-DomainJoin
    Test-ServerManager
)

File: plugins/modules/win_reboot_info.ps1 [hunk 8]
```diff
@@ -0,0 +1,183 @@
...
+$module.Result.pending_reboot = $reasons.Count -gt 0
+$module.Result.pending_reboot_reasons = @($reasons)
```
jborean93: If using the `$reasons = @(...)` above this can just become

```suggestion
$module.Result.pending_reboot_reasons = $reasons
```

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+
+$bootTime = Get-LastBootTime -Module $module
+$module.Result.last_reboot = @{
+    time = $bootTime.last_boot_time
+    time_epoch = $bootTime.last_boot_time_epoch
+    details = Get-LastRebootEvent -Module $module -BootTime $bootTime.utc_datetime
+}
+
+$reasons = Get-PendingRebootReason
+$module.Result.pending_reboot = $reasons.Count -gt 0
```
jborean93: I personally would change this to `reboot_required` to align with how other modules do. I know a pending reboot is somewhat different to required but `reboot_required` seems to be a general common convention. What do you think?

plugins/modules/win_reboot_info.py
```diff
@@ -0,0 +1,119 @@
...
+      - This is gathered from Event ID 1074 in the System log.
+      - Will be C(null) if no shutdown event was found in the event log.
+      type: dict
+      contains:
+        initiated_by:
+          description: The user account that initiated the reboot.
+          type: str
+          sample: "DOMAIN\\admin"
+        process:
+          description: The process that triggered the reboot.
```
jborean93: Does this contain just the executable or the executable + args? I would potentially change this to be `executable`/`process_executable` or `process_command_line` depending on that. Having `process` by itself lends more towards something like a PID to me.

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,183 @@
...
+
+        if ($rebootEvent) {
+            $eventTime = $rebootEvent.TimeCreated.ToUniversalTime()
+            if ($eventTime -lt $BootTime.AddMinutes(-5)) {
+                return $null
+            }
+
+            return @{
+                process = $rebootEvent.Properties[0].Value
+                reason = $rebootEvent.Properties[2].Value
```
jborean93, line 64: We may also want `reason_code = $rebootEvent.Properties[3]` to provide a language agnostic value here and also a proper value if the provided reason code didn't match up with a known reason to Windows.

tests/integration/targets/win_reboot_info/aliases
```diff
@@ -0,0 +1 @@
+shippable/windows/group1
```
jborean93: `group1` has recently been reserved for modules with action plugins as it is run across a range of Ansible versions. As this is just a remote module without any integration on Python/Ansible Controller versions we should target a different group that only tests the remote side.

changelogs/fragments/06082026-win_reboot_info.yml
mikemorency: thats cool

tests/integration/targets/win_reboot_info/aliases
mikemorency: moved it to group 4 with win_regedit, since these tests also rely on that module.

plugins/modules/win_reboot_info.ps1
mikemorency: I think consistency is a good idea; I would have chosen that had I thought to check. Updated to `reboot_required` :+1:

plugins/modules/win_reboot_info.ps1
mikemorency: yea thats a good suggestion, it is definitely simpler

plugins/modules/win_reboot_info.py
```diff
@@ -0,0 +1,119 @@
...
+          type: str
+          sample: "C:\\Windows\\system32\\shutdown.exe"
+        reason:
```
jborean93, line 79: We need to add `reason_code`.

jborean93 (commented):
Tiny documentation issue, otherwise the rest looks great.

plugins/modules/win_reboot_info.py
mikemorency: Can you double check this? I was reading about the reason_code and it made sense to try to return the hex string to the user since that is what they would see in event viewer. https://github.com/ansible-collections/ansible.windows/pull/936/changes#diff-f4c50a653b861d4164810d4d0c78c0d0270df96f121ee7646a9b4d97ca396051R27-R42

jborean93: Ah I see, the raw event properties return it as a string, that's unfortunate.

I would say we should parse this string and return it as an int because that's essentially what it is. It also makes comparisons with other `int` values more consistent.

Because the property value is a string we would have to do something like

```powershell
$reasonCode = $null
$null = [System.Management.Automation.LanguagePrimitives]::TryConvertTo(
    $rebootEvent.Properties[3].Value,
    [int],
    [ref]$reasonCode)
```

If the value isn't a valid int then it's just kept as `$null` but as this is sourced from an API that only accepts an int then I'm not sure when this would ever not be valid.

Even though the value is seen as a hex representation, we should be providing the int value it represents so people can more easily compare it.

```yaml
# Both of these would work
- reason_code == -1
- reason_code == 0xFFFFFFFF
```

Whereas if you returned it as a hex str then you need additional filters and conversions to convert it back to a number.

mikemorency: Ah good point. Int it is

plugins/modules/win_reboot_info.ps1
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function Format-ReasonCode {
+    param($Value)
+
+    # The reason code from event 1074 is returned as a string. Parse it as an integer since
+    # that's what it actually represents, using Int64 rather than Int32 since the high bit
+    # (the "planned" flag, e.g. 0x80040002) makes the value exceed Int32's range. Fall back
+    # to returning the raw value unchanged if it can't be parsed so a malformed field doesn't
+    # drop the rest of the reboot details.
```
jborean93:
```suggestion
    # The reason code from event 1074 is returned as a string. Parse it as an integer since
    # that's what it actually represents, using Int64 rather than Int32 since the signed bit
    # (the "planned" flag, e.g. 0x80040002) makes the value negative whereas YAML
    # parses hex values as unsigned. This allows users to more easily compare the returned
    # value with a hex literal '- res.reason_code == 0x80000000'. Fall back to returning the
    # raw value unchanged if it can't be parsed so a malformed field doesn't drop the rest of
    # the reboot details.
```

jborean93 (approved):

plugins/modules/win_reboot_info.ps1
jborean93: Just a slight change to the comment to reflect that YAML parses hex values as unsigned which is why we do the `long` trick here.

PR #946, opened 2026-09-13
feat(dsc3): support DSC 3.3 directives, what-if and restart reporting

Gijsreyn:
##### SUMMARY

Microsoft DSC 3.3 is around the corner (probably next week GA release), so this pull request brings the `dsc3` module up to date. As the module was written against DSC 3.0, it misses several things that have changed since:

- Configuration document `directives` (`resourceDiscovery`, `securityContext`, `version`) added in DSC 3.2 had no first-class option, and a violated directive only produced `Resource error (ExitCode=2)` because the DSC error is on stderr.
- Group resources (`Microsoft.DSC/Group`, `Microsoft.DSC/Include`) nest their results and DSC reports a group as out of state even when every child is in state, so check mode marked every group as changed.
- The operation was read from `metadata.Microsoft.DSC`, which DSC deprecated in 3.2 in favour of `executionInformation` and removes in 4.0.
- Exit codes 6-10 were unmapped, `remote_config_file` was documented as `true` while the action plugin defaults to `false`, and the action plugin dropped the base action result.
- No way to use `--what-if`, `--parameters-file`, `--system-root`, `--ignore-settings-file`, `DSC_RESOURCE_PATH`, a working directory (controls `Microsoft.DSC/Include` path resolution) or a `dsc.exe` that is not on `PATH`. `_restartRequired` reported by resources was never surfaced.

The PR's design decision was to keep older 3.x releases working. 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

dsc3

##### ADDITIONAL INFORMATION

Verified by running the module through `Ansible.Basic` on Windows PowerShell 5.1 against the `DSC-3.3.0-rc.2` release, and `ansible-test sanity`/`units` (ansible-core 2.21 test container). The integration target still needs a CI run on a Windows host.


Gijsreyn:
Hey @jborean93 - opening this up in an attempt to learn a bit of Ansible. As the description also states, DSC 3.3 is around the corner, so I thought, why not open a PR to keep it up to date. If you've time, do you mind taking a look at it?

jborean93:
Thanks for submitting the PR. I hope to get around to looking at this next week, sorry this week is just a bit busy with a few other things going on. Testing might be a bit difficult as we will want to ensure that we remain compatible with the older versions while also hopefully testing the newer ones where we've added support for these new features.

For the test failures

> "msg": "Task failed: Error while evaluating conditional: object of type 'dict' has no attribute 'invocation'"

By default Ansible does not add the invocation args hence the failure. Typically we don't really need to test these because we are providing these arguments and it's mostly just a reflection of the callsite.

